PR #1, opened 2024-04-22
Modified params_mapping to be able to use AbacusHOD keys by default

SBouchard01:
Also removed the list case from params_mapping as it is only used once and can be handled by exchanging two lines in check_params.


SBouchard01:
Added back list handling to param_mapping for class initialization

PR #17, opened 2025-07-03
Implement power spectrum estimator

epaillas:
This PR extends the PolyBin3D class to include the power spectrum estimator, which supports a jax backend and has the same dependencies as the bispectrum estimator. 

The imports now work as follows:

`from acm.estimators.galaxy_clustering.polybin3d import PowerSpectrum, Bispectrum`


SBouchard01 (approved):
Are we sure we want the estimator module to be called `polybin3d` instead of `bispectrum` ? except for that, all good, no conflicts, can be merged

epaillas:
Both the power spectrum and bispectrum class are now under the "polybin3d" module to make it explicit that those statistics are being calculated with the PolyBin3D set of estimators rather than the Yamamoto estimator implemented in e.g. pypower (estimator from a theory perspective, not just the code). I think this will be a bit cleaner in case we implement more power spectrum/bispectrum estimators in the future.

PR #23, opened 2025-08-13
Format lightcone

hbrincon:
Redesign of the lightcone class to inherit from the cutsky class. Docstrings are also now included for the lightcone class.

### DIFF
--- a/acm/hod/lightcone.py
+++ b/acm/hod/lightcone.py
@@ -115,7 +115,8 @@ class BaseLightconeCatalog(ABC):
             norm = np.linalg.norm(samples, axis=0)
             num_in_lightcone = np.sum((norm>dmin) * (norm<dmax))
             volume = 2000**3 * num_in_lightcone/nsamples_per_box
-            nbar = len(data['Z']) / volume
+            correction = 8 if full_sky else 1  # multiply by 8 if only using the full sky
+            nbar = len(data['Z']) / (volume * correction)
             
         else:
             # Abacus huge (assumes that shells are entirely contained within the survey box)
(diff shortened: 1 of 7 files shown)


acm/hod/lightcone.py
```diff
@@ -411,4 +160,311 @@ def shell_filling_fraction(self, shells):
...
+            xyz_pos = self.catalog['Position']
+            mask = (xyz_pos[:,0] < 2000) * (xyz_pos[:,1] < 4000) * (xyz_pos[:,2] < 4000)
+            mask = mask * ( (xyz_pos[:,1] < 2000) + (xyz_pos[:,2] < 2000) )
+
+            for key in self.catalog.keys():
+                self.catalog[key] = self.catalog[key][mask]
+        
+    def apply_angular_mask(self):
+        # TODO: determine args
+        BaseLightconeCatalog.apply_angular_mask(self)
```
epaillas, line 449: I'm not really sure these re-definitions of methods are needed, unless you are going to change the method implementation itself. The class is meant to inherit from the base class, so you should be able to simply call `self.apply_angular_mask` without defining this function.

 We should also make a call to `super` to initialize the base class, in case the base ends up having some executable code under its own `init` (for the moment, it is just a `pass`). Somewhere in the __init__ of LightconeRandoms, I would add
 
 `super().__init__()`
 
 similar to how it's done here
 https://github.com/epaillas/acm/blob/af73909173043bbb9f1b1ce07ef497127de66e99/acm/observables/emc/density_split_correlation.py#L13

hbrincon: I've added calls to the base class constructor for both the lightcone and cutsky classes.

PR #28, opened 2025-09-22
Cosmodesi bugfix

SBouchard01:
Fixing `.yaml` files not found in non-editable install, `sunbird` call to `sunbird.emulators.train.fit` logger directory argument and removed `sunbird` dependencies that should _in theory_ be handled by `sunbird` install (PR #45 of `sunbird` pending).


hbrincon (approved):
These changes appear straightforward to me. So long as the installation has been tested and works, this should be good to go

PR #33, opened 2025-10-03
Add example notebook on covariance calculations

epaillas:
This PR adds a new notebook with examples on how to compute the covariance matrix from a set of observables.

I noticed the cosmodesi branch is lacking some covariance matrix processing features we have on the harmonized branch. For now, I have only implemented one of them that is necessary to compute the joint covariance of statistics which do not share the same test set. I will implement the other features in the near future.


SBouchard01:
I'll also add the changes to the covariance stuff to PR #27 (this will probably create merge conflicts), unless we ditch this one, merge PR #27 to `cosmodesi` and then add the covariance stuff back 🤔

PR #40, opened 2025-10-18
Spherical void statistic and general bug fixes

ntbfin00:
I have implemented the spherical voidfinder from my code [VERSUS](https://github.com/ntbfin00/VERSUS). Ideally it would've been nice to merge this with the recent jax handling of mesh-based statistics but the code requires that cells are assigned galaxies by 'nearest-grid-particle' not by  'cloud-in-cell'. I think this is not straightforward with `jaxpower` and would require more thought.

Also I fixed a bug in the `pyproject.toml` that was preventing me from installing the ACM package and fixed some other minor bugs I spotted in the `abacus_base.py` file.

### DIFF
--- a/acm/estimators/galaxy_clustering/base.py
+++ b/acm/estimators/galaxy_clustering/base.py
@@ -1,8 +1,7 @@
-from jaxpower import MeshAttrs, ParticleField, FKPField, ComplexMeshField, RealMeshField
+from jaxpower import MeshAttrs, ParticleField, FKPField, ComplexMeshField, RealMeshField, get_mesh_attrs
 from jax import numpy as jnp
 import numpy as np
 import time
-import jax
 
 
 class BaseEstimator:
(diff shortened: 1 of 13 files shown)


epaillas:
This looks good. I solved a small merge conflict. I'm going to test the script and merge the PR later today.

Regarding the CIC interpolation with jaxpower, you can actually use NGP and other interpolation schemes too, so you might want to loop into it in the future:

https://github.com/adematti/jax-power/blob/551edce48bddcaabb4165ad03b181728ccc905ec/jaxpower/mesh.py#L1296

SBouchard01:
the `pyproject.toml` bug will be fixed in the `cosmodesi-bugfix` branch along with some other bugs, thanks for catching it ! :)

epaillas:
I just tested the code and it runs well (and fast!).

PR #47, opened 2025-10-22
Accepting parameters as a dict for model predictions

### LINKED ISSUE #46: Accepting parameters as a dict for model predictions
It would be great if the observables get_model_prediction method could have the capability of accepting the "x" input features as a dictionary, with keys being the parameter names:

https://github.com/epaillas/acm/blob/c7950fe3e62d2b50389d442b2803b9db6e3d0404/acm/observables/base.py#L373

The class would internally figure out the correct ordering of the parameters that needs to be fed to the neural network.

This would be useful in many different instances, but especially when dealing with the results of model fits, when we need to read chain data and pass it to the model prediction to make interpretation plots, estimate the goodness-of-fit, etc.

epaillas:
Resolves #46 

Here's a code snippet to test this:

```
import acm.observables.emc as emc
from sunbird.inference.samples import Chain
import matplotlib.pyplot as plt


observable = emc.ReconstructedGalaxyPowerSpectrumMultipoles(
    paths={
        'data_dir': '/pscratch/sd/e/epaillas/emc/v1.2/abacus/compressed/',
        'measurements_dir': '/pscratch/sd/e/epaillas/emc/v1.2/abacus/',
    },
    numpy_output=True,
)

observable.select_filters = {'cosmo_idx': 0, 'hod_idx': 47, 'multipoles': [0, 2, 4]}

chain_fn = '/global/cfs/cdirs/desicollab/users/epaillas/acm/fits_emc/abacus/oct21/c000_hod047/cosmo-base_hod-base-VB-AB-CB-s/chain_recon_spectrum.npy'
chain = Chain.load(chain_fn)
maxp = chain.data['samples'][chain.data['log_posterior'].argmax()]
maxp = {key: val for key, val in zip(chain.data['names'], maxp)}
maxp.update(chain.data['fixed_parameters'])

print(maxp)
print(observable.get_model_prediction(maxp))
```

### DIFF
--- a/acm/observables/base.py
+++ b/acm/observables/base.py
@@ -376,7 +376,7 @@ class Observable():
         
         Parameters
         ----------
-        x : array_like
+        x : array_like, dict
             Input features.
         model : FCN
             Trained theory model. If None, the model attribute of the class is used. Defaults to None.


acm/observables/emc/base.py
```diff
@@ -168,6 +168,14 @@ def get_model_prediction(self, x, model=None, coords=None, attrs=None, nofilters
         xarray.DataArray | np.ndarray
             Model prediction.
         """
+        if isinstance(x, dict):
+            missing = set(self.x_names) - set(x.keys())
+            assert not missing, (
+                "Input x dictionary keys do not match the model input names. "
+                f"Missing keys: {missing}"
+            )
```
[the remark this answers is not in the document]
SBouchard01: Suggestion on assertion seems relevant, maybe we should check the rest of the classes for assertions too ? (I don't think we have a lot of them, but if they can be ignored, we should add them to the code with that functionality in mind and raise errors instead if we don't want them to be overriden at all)

acm/observables/emc/base.py
```diff
@@ -168,6 +169,21 @@ def get_model_prediction(self, x, model=None, coords=None, attrs=None, nofilters
...
+        if isinstance(x, dict):
+            missing = set(self.x_names) - set(x.keys())
+            extra = set(x.keys()) - set(self.x_names)
+            if missing:
+                raise ValueError(
+                    "Input x dictionary keys do not match the model input names. "
+                    f"Missing keys: {missing}"
+                )
+            if extra:
+                warnings.warn(
```
SBouchard01: Using` logger.warning` will log the warning and also output it as a warning (and there already is a logger in the class ;) )

SBouchard01: Also, do we want this change implemented in the EMC observable classes, or overall in the base Observable  class to have this option on all projects ? 🤔

Issue #58, opened 2025-10-27
Change BoxHOD output


SBouchard01:
To allow debugging and easier positions & boxsize recovery, we should store only the initial information (`X`, `Y`, `Z`, `VX`, `VY`, `VZ`, `qpar`, `qperp`) and recover any derived information from this trough class methods

e.g. `boxsize` and RSD/AP positions should be recovered through a `get_positions` & `get_boxsize` method (ideally a static method that takes the output dict as an input to be abe to use it after saving the file

Related to #57 second issue

PR #59, opened 2025-10-27
Cosmodesi box outputs

SBouchard01:
Sets the RSD and AP distorsions application outside of the `run()` method, to allow:
- storing only the information that allows the reconstruction of the different elements we want to compute
- on-the-fly computation of the positions and boxsizes depending on the los

> [!Warning]
> This will need to be taken into account when passing `BoxHOD` outputs to `CutskyHOD` !
> (We don't pass RSD or AP distorsions ATM, but if we do we'll need to be careful about this or need to implement similar methods to the `CustkyHOD` class)

> [!Important]
> Please, merge #57 before merging this one, it will cause merge issues but it will be simpler to solve them in this way than the other

> Solves #58

### DIFF
--- a/acm/hod/box.py
+++ b/acm/hod/box.py
@@ -452,6 +452,7 @@ class BoxHOD:
         np.ndarray
             Array of galaxy positions with shape (N_gal, 3).
         """
+        hod_dict = hod_dict.copy()  # Avoid modifying the original dictionary
         tracer_dict = hod_dict[tracer] if tracer is not None else hod_dict
         
         # Apply RSD before AP distortions
@@ -499,6 +500,7 @@ class BoxHOD:
         -------
         dict
             Dictionary containing the HOD catalog with redshift-space distortions to the specified axis.
+            Will overwrite the input `tracer_dict` in place, use a copy if needed !
         """
         ax = los.upper()
         offset = boxsize / 2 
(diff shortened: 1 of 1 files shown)


SBouchard01:
## Example :

For accessing the box positions with RSD & AP on 3 line of sight:
```python 
box = BoxHOD(...)
hod_dict = box.run(...)['LRG'] # Now saves only X, Y, Z, VX, VY, VZ + new header info
for los in ['x', 'y', 'z']:
    boxsize = BoxHOD.get_boxsize(
        boxsize=box.boxsize, 
        add_ap=box.add_ap, 
        los=los, 
        q_par=box.q_par, 
        q_perp=box.q_perp
    ) # Returns 3 element array or float if add_ap is False
    positions = BoxHOD.get_positions(
        hod_dict, 
        los=los, 
        add_rsd=True, 
        hubble=box.hubble, 
        az=box.az, 
        boxsize=box.boxsize,
        add_ap=box.add_ap # Can also be set to True by default
        q_par=box.q_par,
        q_perp=box.q_perp,
    ) # Returns array of size (Ngal, 3)
```

You can also get this from the new galaxy catalogs (if saved with this version) : 
```python
import fitsio
catalog, header = fitsio.read(save_fn, header=True)
boxsize = BoxHOD.get_boxsize(
    header['BOXSIZE'], # casing doesn't matter
    add_ap=True, # Depending on what you want to measure
    los='x', 
    q_par=header.get('Q_PAR', 1.0), 
    q_perp=header.get('Q_PERP', 1.0),
)
positions = BoxHOD.get_positions(
      catalog, 
      los='x', 
      add_rsd=True, # Depending on what you want to measure
      hubble=header.get('hubble', None), # Setting to None by default for the example
      az=header.get('az', None),
      boxsize=header.get('boxsize', None),
      add_ap=True # Depending on what you want to measure
      q_par=header.get('q_par', None),
      q_perp=header.get('q_perp', None),
)
```

PR #75, opened 2025-11-03
Implement custom emulator covariance for (joint) observables and fix a few plotting bugs

epaillas:
I have implemented a few options to calculate the emulator covariance for single statistics (mean absolute deviation, median absolute deviation, standard deviation), both in diagonal and non-diagonal versions. I have also enabled a joint block-diagonal covariance of multiple statistics. The non-block-diagonal cannot be imlpemented yet as some statistics still have fewer measurements than others.

I have tested this at the parameter inference level and things seem to be working as expected.

### DIFF
--- a/scripts/emc/inference/inference_abacus.py
+++ b/scripts/emc/inference/inference_abacus.py
@@ -1,0 +1,237 @@
+from sunbird.inference.pocomc import PocoMCSampler
+from sunbird.inference import priors as sunbird_priors
+from sunbird.cosmology.model_params import get_model_params
+
+import acm.observables.emc as emc
+from acm.observables import CombinedObservable
+from acm.utils.covariance import get_covariance_correction
+from acm import setup_logging
+
+from cosmoprimo import fiducial
+
+from pathlib import Path
+import numpy as np
+import argparse
+import logging
+
+
+def get_priors(cosmo=True, hod=True):
+    """
+    Return a dictionary of prior distributions, hard limits (ranges),
+    and labels for cosmological and HOD parameters in a format
+    that is readable by PocoMCSampler.
+    """
+    stats_module = 'scipy.stats'
+    priors, ranges, labels = {}, {}, {}
+    if cosmo:
+        priors.update(sunbird_priors.AbacusSummit(stats_module).priors)
+        ranges.update(sunbird_priors.AbacusSummit(stats_module).ranges)
+        labels.update(sunbird_priors.AbacusSummit(stats_module).labels)
+    if hod:
+        priors.update(sunbird_priors.Yuan23(stats_module).priors)
+        ranges.update(sunbird_priors.Yuan23(stats_module).ranges)
+        labels.update(sunbird_priors.Yuan23(stats_module).labels)
+    return priors, ranges, labels
+
+
+def get_fixed_params(cosmo_model, hod_model):
+    """
+    Return a list of fixed parameter names based on the cosmological and HOD models.
+    This function checks which parameters are free in the specified models.
+    """
+    free = []
+    # cosmology
+    if 'base' in cosmo_model:
+        free += ['omega_b', 'omega_cdm', 'sigma8_m', 'n_s']
+    if 'w0' in cosmo_model:
+        free += ['w0_fld']
+    if 'wa' in cosmo_model:
+        free += ['wa_fld']
+    if 'Nur' in cosmo_model:
+        free += ['N_ur']
+    if 'nrun' in cosmo_model:
+        free += ['nrun']
+    if 'fixed-ns' in cosmo_model:
+        free.remove('n_s')
+    # HOD
+    if 'base' in hod_model:
+        free += ['logM_cut', 'logM_1', 'sigma', 'alpha', 'kappa']
+    if 'AB' in hod_model:
+        free += ['B_cen', 'B_sat']
+    if "CB" in hod_model:
+        free += ["A_cen", "A_sat"]
+    if 'VB' in hod_model:
+        free += ['alpha_c', 'alpha_s']
+    if '_s' in hod_model or '-s' in hod_model:
+        free += ['s']
+    fixed = [par for par in priors.keys() if par not in free]
+    return fixed
+
+def get_filters(observable_name):
+    """
+    Get the select and slice coordinates for the observable.
+    This function returns dictionaries that specify which coordinates to select
+    and which to slice for the given observable.
+    """
+    select_filters = {'cosmo_idx': args.cosmo_idx, 'hod_idx': args.hod_idx}
+    slice_filters = {}
+    """Get the select and slice coordinates for the observable."""
+    if observable_name == 'GalaxyCorrelationFunctionMultipoles':
+        select_filters.update({'multipoles': [0, 2]})
+        slice_filters.update({'s': [0.0, 150]})
+    elif observable_name == 'GalaxyPowerSpectrumMultipoles':
+        select_filters.update({'multipoles': [0, 2, 4]})
+    elif observable_name == 'GalaxyBispectrumMultipoles':
+        select_filters.update({'multipoles': [0, 2]})
+        slice_filters.update({'k': [0.0, 0.7]})
+    elif observable_name == 'ReconstructedGalaxyPowerSpectrumMultipoles':
+        select_filters.update({'multipoles': [0, 2, 4]})
+        slice_filters.update({'k': [0.0, 0.7]})
+    elif observable_name == 'DensitySplitPowerSpectrumMultipoles':
+        select_filters.update({'statistics': ['quantile_data_power']})
+    return select_filters, slice_filters
+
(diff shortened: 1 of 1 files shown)


acm/observables/emc/density_split_correlation.py
```diff
@@ -2,32 +2,34 @@
...
-class DensitySplitCorrelationFunctionMultipoles(BaseObservableEMC):
+class DensitySplitGalaxyCorrelationFunctionMultipoles(BaseObservableEMC):
     """
     Class for the Emulator's Mock Challenge density-split correlation
     function multipoles.
     """
     def __init__(self, **kwargs):
-        super().__init__(stat_name='dsc_conf', **kwargs)
-        self.n_test = 6*100 #Override default number of test samples 
+        super().__init__(stat_name='ds_xiqg', n_test=6*3, **kwargs)
```
SBouchard01: Suggestion: `n_test` could be passed as an argument of __init__ to allow overriding if needed ?

```suggestion
    def __init__(self, n_test=6*3, **kwargs):
        super().__init__(stat_name='ds_xiqg', n_test=n_test, **kwargs)
```

> (Also valid for other EMC Observables)

acm/observables/emc/density_split_correlation.py
```diff
@@ -146,41 +145,40 @@ def compress_data(
             Compressed dataset containing 'x' and 'y' DataArrays.
             If add_covariance is True, also contains 'covariance_y' DataArray.
         """
-        base_dir = self.paths['statistic_dir']
+        base_dir = Path(self.paths['measurements_dir'],  f'base/density_split/')
```
SBouchard01:
```suggestion
        base_dir = Path(self.paths['measurements_dir']) / 'base' / 'density_split'
```
> [nitpick] for consistency

acm/observables/base.py
```diff
@@ -485,24 +486,171 @@ def get_covariance_matrix(self, volume_factor: float = 64, prefactor: float = 1)
...
+            if diag:
+                mad = median_abs_deviation(cov_y, axis=0)
+                mad *= 1.4826  #  make summary consistent with stdev for a normal distribution
+                cov = np.diag(mad ** 2)
+            else:
+                cov = self.orthogonal_gk_mad_covariance(cov_y.values)
+        elif method == 'mean':
+            if diag:
+                mad = np.mean(np.abs(cov_y - np.mean(cov_y, axis=0)), axis=0)
+                mad *= 1.2533  # make summary consistent with stdev for a normal distribution
```
SBouchard01: What is this number for, and is it OK to hardcode it ? (and if yes, is this an exact value or an approximation ?)

acm/observables/combined.py
```diff
@@ -207,15 +210,34 @@ def get_covariance_matrix(
...
+    def get_emulator_covariance_matrix(self, prefactor: float = 1, method: str = 'median', diag: bool = False) -> np.ndarray:
         """
-        cov_y = self.emulator_covariance_y
-        prefactor = prefactor
-        
-        cov = prefactor * np.cov(cov_y, rowvar=False)
-        return cov
+        Covariance matrix of the emulator residuals for a combination of multiple summary statistics.
+        The matrix is block-diagonal, with each block corresponding to the covariance matrix of each
+        individual observable.
```
SBouchard01: Removing cross-covariance terms on purpose an enforcing block-diagonal emulator covariance matrix: 
- Why is there no option to have the full covariance matrix anymore ? (just to understand)
- Should we enforce this for `get_covariance_matrix` too ?

acm/observables/base.py
```diff
@@ -485,24 +486,171 @@ def get_covariance_matrix(self, volume_factor: float = 64, prefactor: float = 1)
...
             cov_y = cov_y.reshape(cov_y.shape[0], -1) # Expect first dimension to be the sample dimension
         elif len(cov_y.shape) < 2:
             self.logger.error("Covariance array has less than 2 dimensions, covariance matrix computation might return some unexpected results.")
         
-        cov = prefactor * np.cov(cov_y, rowvar=False)
-        return cov
+        if method == 'median':
+            if diag:
+                mad = median_abs_deviation(cov_y, axis=0)
+                mad *= 1.4826  #  make summary consistent with stdev for a normal distribution
```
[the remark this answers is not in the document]
epaillas: See https://en.wikipedia.org/wiki/Median_absolute_deviation 

(the paragraph starting with "For normally distributed data" ).

The MAD is used in this context as a more robust summary of the dispersion in the data that is less sensitive to outliers (in our case, edge of our priors that give crazy emulator error). However, even if the residuals did not have crazy outliers and were nice and Gaussian, the MAD would still be a factor of ~1.48 smaller than the standard deviation, which we don't want, so this corrects for it. It is an approximation to the expression that involves the quantile function shown in Wikipedia.

acm/observables/base.py
```diff
@@ -485,24 +486,171 @@ def get_covariance_matrix(self, volume_factor: float = 64, prefactor: float = 1)
...
+        # cov(X,Y) ≈ (1/4)[ Var(X+Y) - Var(X-Y) ], with Var estimated by MAD^2.
+        C = np.empty((n_bins, n_bins), dtype=X.dtype)
+        for j in range(n_bins):
+            C[j, j] = s2[j]
+            for k in range(j+1, n_bins):
+                sp = _mad_1d(X[:, j] + X[:, k])**2
+                sm = _mad_1d(X[:, j] - X[:, k])**2
+                cov_jk = 0.25 * (sp - sm)
+                C[j, k] = C[k, j] = cov_jk
+        return C
```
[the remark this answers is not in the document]
SBouchard01: Could you also move the related covariance functions from `acm.utils.tools` there while you are at it ? :) (I think that's all the functions in the files TBH)
It would be more coherent with the module name (and I don't think any of our `acm` files calls it, I made this for the scripts that call the covariance from observables)

acm/observables/base.py
epaillas: [d78c527](https://github.com/epaillas/acm/pull/75/commits/d78c5270b55d1f167b7827a632c6c83edacee19c) now implements exact expressions for both cases

SBouchard01: Great ! :) 
There is also one in `mad_1d` that could be also changed too ^^

acm/observables/combined.py
SBouchard01: That's a good plan ! 
And what about `CombinedObservable.get_covariance_matrix` (for the mocks), do we implement a similar behavior for consistency, or do we keep the full covariance ? (this question also applied to `Observable.get_covariance_matrix`, as we didn't implement the different methods there)

acm/observables/base.py
```diff
@@ -485,25 +487,65 @@ def get_covariance_matrix(self, volume_factor: float = 64, prefactor: float = 1)
...
             cov_y = cov_y.reshape(cov_y.shape[0], -1) # Expect first dimension to be the sample dimension
         elif len(cov_y.shape) < 2:
             self.logger.error("Covariance array has less than 2 dimensions, covariance matrix computation might return some unexpected results.")
-        
-        cov = prefactor * np.cov(cov_y, rowvar=False)
-        return cov
-   
+
+        if isinstance(cov_y, xarray.DataArray):
+            cov_y = cov_y.values
```
SBouchard01: This can be moved line 514 as it falls under the same if statement

acm/observables/base.py
```diff
@@ -485,25 +487,65 @@ def get_covariance_matrix(self, volume_factor: float = 64, prefactor: float = 1)
...
+            else:
+                raise NotImplementedError("Mean absolute deviation covariance is not implemented for full matrix (diag=False).")
+        elif method == 'stdev':
+            if diag:
+                std = np.std(cov_y, axis=0)
+                cov = np.diag(std ** 2)
+            else:
+                cov = np.cov(cov_y, rowvar=False)
+        else:
+            raise ValueError(f"Unknown method '{method}' for emulator covariance matrix computation.")
```
SBouchard01: **Only if we decide to implement similar covariance computation on other methods (e.g. `get_covariance_matrix`):** This could be moved in an external `covariance_matrix_from_array(cov_y, method, diag) -> cov` external function in `acm.utils.covariance`, to ensure a consistent behavior with similar options and avoid copying code ^^

*Otherwise, ignore this comment :)*

acm/observables/combined.py
SBouchard01: Also, see [this comment](https://github.com/epaillas/acm/pull/75/files#r2492117346) if you plan to add a covariance computation method argument ^^

epaillas: I think these different methods are needed to understand the emulator error, which can be ill-behaved and there is no unique standard definition in the literature for it, but I wouldn't use them for the data covariance. The way to estimate it is pretty well established and the noise coming from the different mock realizations should be fairly Gaussian. If it's not, then the assumption of a Gaussian likelihood at the parameter inference level needs to be revisited.

PR #96, opened 2025-11-21
Fix bugs in MST implementation and non-cubic box capability

knaidoo29:
Fixed bugs in MST implementation to be able to run with the latest ACM pipeline and added functionality to use with non-cubic boxes.

### DIFF
--- a/acm/estimators/galaxy_clustering/mst.py
+++ b/acm/estimators/galaxy_clustering/mst.py
@@ -3,7 +3,7 @@
 import numpy as np
 import logging
 import time
-from .base import BaseDensityMeshEstimator
+from acm.estimators.galaxy_clustering.base import BaseDensityMeshEstimator
 
 
 class MinimumSpanningTree(BaseDensityMeshEstimator):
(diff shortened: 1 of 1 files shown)


acm/estimators/galaxy_clustering/mst.py
```diff
@@ -220,17 +234,18 @@ def get_percolation_statistics(self, data_pos):
...
 
-        # Normalize by the volume
-        mstdict['mst1pt'] /= self.boxsize**3
+        Normalize by the volume
+        mstdict['mst1pt'] /= self.boxsize[0]*self.boxsize[1]*self.boxsize[2]
         for N in range(2, self.Nthpoint+1):
-            mstdict['mst%ipt'%N] /= self.boxsize**3
-            mstdict['end%ipt'%N] /= self.boxsize**3
+            mstdict['mst%ipt'%N] /= self.boxsize[0]*self.boxsize[1]*self.boxsize[2]
+            mstdict['end%ipt'%N] /= self.boxsize[0]*self.boxsize[1]*self.boxsize[2]
```
SBouchard01: `np.prod(self.boxsize)` could be used on `self.boxsize` instead of reading all elements ?

knaidoo29:
Thanks @SBouchard01 , I deleted the branch because I found a typo but let me re-post with those changes you suggested implemented

PR #97, opened 2025-11-21
Implemention MST with non-cubic boxes

knaidoo29:
* fixes some bugs in the MST implementation and some incompatibility with the new acm pipeline.
* implements some changes suggested by @SBouchard01 
* implements functionality with non-cubic boxes


epaillas:
This looks great, thanks for implementing it!

Have you tested that it runs with the current version of the pipeline? Can we add an MST function here to measure it on the new EMC mocks?

https://github.com/epaillas/acm/blob/a0ba7bf99fcb4f79adb92172c2e8b462c20cb733/scripts/emc/measurements/abacus_base.py#L57

I'm still fine with you measurement things independently, but it would be nice to check that the pipeline output matches your output for at least a couple mocks.

Issue #98, opened 2025-11-21
Adding training / validation / testing sub-datasets within the Observables datasets


epaillas:
We have gotten a request to include information about the training / validation / test set as part of the compressed datasets. If that split is already decided by the time the files are being compressed, we could just add some entries here:

https://github.com/epaillas/acm/blob/a0ba7bf99fcb4f79adb92172c2e8b462c20cb733/acm/observables/emc/spectrum.py#L183

so that apart from 'x' we also have 'x_train', 'x_val', 'x_test', etc, which would be subsets of 'x'. Same thing for 'y'.

@PalPaulD I will try to implement an example for the galaxy power spectrum and then you can tell me if that's sufficient.

SBouchard01:
I can make a quick proof of concept now and take a closer look on Monday :)

epaillas:
(Also taking into account that we might have more HOD samples in the test set tham we do in the training set... this is definitely happening for some statistics already, but I haven't actually implemented that in terms of code because we don't have that functionality yet. Having something like obs.x_test would make it easy I think).

SBouchard01:
As a note, I thought `sunbird` was handling the split between training and validation set? Do we still need to split this one ourselves?

SBouchard01:
PR merged, I'll make a pass on the EMC observables soon to add the train/test split in the compression methods, as specified in #106

PR #102, opened 2025-11-24
Cosmodesi train split

SBouchard01:
Add `split_vars` utils and `drop_nan_dimensions` Observable method to drop the NaN dimensions due to the test/train filters if specified.

Most of the files are changed because I renamed `acm.utils.xarray_data` to `acm.utils.xarray`.

I have provided an example on how to integrate the test/train split to the data compression in acm/observables/emc/tpcf.py, please propagate this to the other Observables !

> [!NOTE]
> Once this PR is merged, I will open a BGS PR to `cosmodesi` (that contains the BGS scripts & Observables). Please merge this PR before the BGS one to allow me to handle the merge conflicts properly 👍🏼 

Should solve #98

### DIFF
--- a/acm/utils/xarray.py
+++ b/acm/utils/xarray.py
@@ -70,5 +70,4 @@ def split_vars(*data_vars, **kwargs):
     for v in data_vars:
         v_in = v.sel(**kwargs)
         v_out = v.drop_sel(**kwargs)
-    
         yield v_in, v_out


SBouchard01:
> Shouldn't we have trian/val split included here too? The purpose of this split at preprocessing level is to be able to make some decisions based on train data before doing any training. Then if a training code decides to split this training set into train+val sets, this will lead to leakage: val data will implicitly inform preprocessing.
> 
> Also fixing train/val split once is healthier for emulators performance comparison.

I don't know, I'm not as involved (yet) in the `sunbird` architecture, but I remember the train/validation split being set by the `val_fraction` parameter in `sunbird.data.ArrayDataModule` (see [here](https://github.com/florpi/sunbird/blob/b2ae777882d971bdc50eaca9c92fc9aa9f39a6c6/sunbird/data/data.py#L39)): It chooses a random fraction of the training set to be removed
Since this is how the training pipeline is designed, I did not add the validation split (but if you want to, you could by either using the same method or another loop on `split_vars`)

I guess this is a discussion to have on how we want `sunbird` to behave - we should definitely bring this up if you think it's relevant ^^

PR #125, opened 2025-12-17
Fix densitysplit class import in EMC observables (forgotten in #118)

SBouchard01:
Everything else seems to work fine, if here are any other bugs let me know I can add them to this PR


SBouchard01:
> [!WARNING]
> I merged the PR without proper review, but this is a very small change it should not crash the code

PR #129, opened 2026-01-05
Arrays in cython in python3.12

PalPaulD:
The code stopped compiling due to some changes in cython. The arrays types must be specified differently now, this is a small patch that adjusts for this. No functional changes, but now acm package compiles in new cosmodesi env


SBouchard01:
Thank you very much ! 
If it compiles and behaves as expected, could you also remove the commented part ? (We can always revert to a previous version trough `git` if needed ^^)

I also notice the `acm/estimators/galaxy_clustering/src/minkowski.pyx` file also has the same formatting, does it breaks the compilation too? (otherwise I'm opening an issue for a later fix)

PR #144, opened 2026-01-27
Rework backend of estimator classes

epaillas:
This PR offers a new implementation that better supports different backends for the estimator class. The different backends (jaxpower and pypower) live in separate files and have their own imports. The user should only need to specify the desired backend when running their statistic.

```
for backend in ['jaxpower', 'pypower']:
    box_args = get_box_args(boxsize, cellsize=100)
    wst = WaveletScatteringTransform(data_positions=positions, backend=backend, **box_args)
    wst.set_density_contrast()
    coeff = wst.run()
```

I have checked that these two backends agree at the density contrast and coefficients level. So far only validated for WST, hence why it's just a draft for now.

A more complete example can be found in the `tests/` directory at the root.

### DIFF
--- a/acm/estimators/galaxy_clustering/base.py
+++ b/acm/estimators/galaxy_clustering/base.py
@@ -1,8 +1,12 @@
+from typing import Any
+import numpy.typing as npt
+
+
 class BaseEstimator:
     """
     Base estimator class.
     """
-    def __init__(self, backend='jaxpower', **kwargs):
+    def __init__(self, backend: str = 'jaxpower', **kwargs: Any) -> None:
         self.backend_name = backend
         
         # Lazy import of backend classes to avoid forcing installation of all backends
@@ -21,7 +25,7 @@ class BaseEstimator:
         else:
             raise ValueError(f"Unknown backend '{backend}'. Available backends: 'jaxpower', 'pypower', 'pyrecon'")
         
-    def read_density_contrast(self, positions, resampler='cic'):
+    def read_density_contrast(self, positions: npt.NDArray, resampler: str = 'cic') -> npt.NDArray:
         """
         Get the density contrast at the input positions.
 z
@@ -47,7 +51,7 @@ z
                 raise NotImplementedError('Pyrecon backend only supports CIC resampling.')
             return self.backend.delta_mesh.read_cic(positions)
 
-    def __getattr__(self, name):
+    def __getattr__(self, name: str) -> Any:
         """
         Delegate attribute access to the backend.
 
--- a/acm/estimators/galaxy_clustering/spectrum.py
+++ b/acm/estimators/galaxy_clustering/spectrum.py
@@ -6,6 +6,7 @@ from jaxpower import (
 )
 import jax
 import logging
+from typing import Any, Optional, Union
 from .base import BaseEstimator
 
 
@@ -15,7 +16,7 @@ class PowerSpectrumMultipoles(BaseEstimator):
     https://github.com/adematti/jax-power/
     """
     
-    def __init__(self, **kwargs):
+    def __init__(self, **kwargs: Any) -> None:
         self.logger = logging.getLogger(__name__)
         kwargs.setdefault('backend', 'jaxpower')
         super().__init__(**kwargs)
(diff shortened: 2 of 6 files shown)


epaillas:
wst_backend_comparison

Things seem to be working

epaillas:
Ok, I'm finished for now. I'd like to merge it if you're fine with it @SBouchard01 .

scripts/emc/measurements/abacus_base.py
```diff
@@ -96,9 +95,6 @@ def compute_spectrum_acm(output_fn, ells=(0, 2, 4), los='z', **attrs):
         print(f'Power spectrum (ACM) done in {t1 - t0:.2f} s.')
 
 def compute_bispectrum(output_fn, positions, basis='scoccimarro', los='z', **attrs):
-=======
-def compute_bispectrum(output_fn, positions, basis='scoccimarro', los='z', bin=None, **attrs):
->>>>>>> a0ba7bf99fcb4f79adb92172c2e8b462c20cb733
     from jaxpower import (ParticleField, FKPField, compute_fkp3_normalization, compute_fkp3_shotnoise, BinMesh3SpectrumPoles, get_mesh_attrs, compute_mesh3_spectrum, MeshAttrs)
```
SBouchard01: PEP8

scripts/emc/measurements/abacus_base.py
```diff
@@ -243,43 +238,22 @@ def compute_density_split(output_fn, positions, smoothing_radius=10, ells=(0, 2,
...
-    np.save(output_fn['xiqg'], ccf)
-    print(f'Saving {output_fn["xiqg"]}')
-    np.save(output_fn['xiqq'], acf)
-    print(f'Saving {output_fn["xiqq"]}')
->>>>>>> a0ba7bf99fcb4f79adb92172c2e8b462c20cb733
 
 def compute_wst(output_fn, positions, init=None, **attrs):
     """Compute the wavelet scattering transform using the ACM package."""
     from acm.estimators.galaxy_clustering.wst import WaveletScatteringTransform
     import warnings
```
SBouchard01: PEP8

acm/estimators/galaxy_clustering/spectrum.py
```diff
@@ -6,17 +6,20 @@
...
     """
     Calculate the power spectrum multipoles using jaxpower.
     https://github.com/adematti/jax-power/
     """
     
     def __init__(self, **kwargs):
         self.logger = logging.getLogger(__name__)
+        if 'backend' in kwargs:
+            if kwargs['backend'] != 'jaxpower':
+                raise ValueError("PowerSpectrumMultipoles only supports the 'jaxpower' backend.")
```
SBouchard01: Needs to ensure the backend is indeed `jaxpower` (if it changes in the BaseEstimator class, the error will not be raised)

I suggest checking the backend and raising the error after `super().__init__()` call

acm/estimators/galaxy_clustering/base.py
```diff
@@ -1,433 +1,62 @@
...
+        # Lazy import of backend classes to avoid forcing installation of all backends
+        if backend == 'jaxpower':
+            from .backends.jaxpower import JaxpowerBackend
+            self.backend = JaxpowerBackend(**kwargs)
+        elif backend == 'pypower':
+            from .backends.pypower import PypowerBackend
+            self.backend = PypowerBackend(**kwargs)
+        elif backend == 'pyrecon':
+            from .backends.pyrecon import PyreconBackend
+            self.backend = PyreconBackend(**kwargs)
```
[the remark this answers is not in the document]
SBouchard01: @epaillas digging this up really quickly as I feel there would be a simpler implementation (and my previous comment on this was mostly because there was no lazy loading) :
- Either there is a lazy loading of the backends and it makes sense to check the name of the backend as it was initially done 
- Or there is a global import at the beginning of the file and it would make sense to just check the backend type (as they are all imported) with `isinstance(self.backend, JaxpowerBackend)` 

From what I read on good practices lazy imports, they are mostly relevant when some import costs a lot or if it's specific to the environment. In our case, the pyproject imposes all dependencies to be loaded so implementation 2 makes sense, _unless_ we decide we split the backends in optional dependencies, in which case the first implementation is more relevant as the dependencies can not be all installed.

PR #146, opened 2026-01-27
Modifications to VERSUS measurements

ntbfin00:
- Added manual removal of largest radii bin to obtain clean voids
- No longer performing void resizing to improve VSF smoothness
- Changed overlap condition to no overlapping centres to improve speed and S/N

### DIFF
--- a/acm/observables/emc/versus.py
+++ b/acm/observables/emc/versus.py
@@ -1,10 +1,13 @@
 import xarray
 import numpy as np
-import glob
 from pathlib import Path
-from .base import BaseObservableEMC
+
+import xarray
+import numpy as np
 import matplotlib.pyplot as plt
 from pycorr import TwoPointCorrelationFunction
+
+from .base import BaseObservableEMC
 from acm.utils.default import cosmo_list # List of cosmologies in AbacusSummit
 from acm.utils.plotting import set_plot_style
 from acm.utils.decorators import temporary_class_state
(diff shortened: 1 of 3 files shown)


SBouchard01:
While you are making some changes to the VERSUS measurements, could you make sure the VERSUS observables match the changes made to the Observable classes in #141 ? (I did not change those as I wasn't sure what could break and I could not test it yet)
Thank you ? 🙏

acm/observables/emc/versus.py
```diff
@@ -111,22 +128,25 @@ def compress_data(
...
+        logger = cls.get_logger()
+        
+        base_dir = Path(paths['measurements_dir']) / 'base' / '{}spherical_voids'.format('recon_' if cls.recon else '')
 
         y = []
         hods = {}
         for cosmo_idx in cosmos:
-            self.logger.info(f'Compressing c{cosmo_idx:03d}')
-            handle = f'c{cosmo_idx:03d}_ph{phase:03d}/seed{seed}/sv_{self.label}_c{cosmo_idx:03d}_hod*.npy'
+            logger.info(f'Compressing c{cosmo_idx:03d}')
```
SBouchard01: Depending on the number of cosmologies, this might be a bit noisy in the info loglevel, maybe move it to debug ? (also the number of cosmologies will be contained in the xarray shape logged below)

acm/observables/base.py
```diff
@@ -108,7 +108,7 @@ def __init__(
         if model is not None:
             self.model = model
         # Try to load model if not provided
-        elif paths is not None and 'model_dir' in paths:
+        elif paths is not None:
```
SBouchard01: good catch ! :)

PR #149, opened 2026-01-30
Update dependency logic and improve #144 Estimators typing

SBouchard01:
Since #144 added lazy imports for the `BaseEstimator` backends in order to avoid having to install all the dependencies, I moved those potentially non-installed dependencies from the `cosmodesi` option to their own.

> [!TIP]
> See [this message](https://github.com/epaillas/acm/pull/144#discussion_r2742840911) to understand the logic behind this change

I also changed a few type hints (e.g. `Any` or `Optional[Any]` is the default behavior and is not necessary to provide)
> Also, if I'm nitpicking a bit, there _is_ a semantic difference between `npt.NDArray` and `npt.ArrayLike`, but I ended up not changing that because honestly it's not very important to what we are doing here 😅


SBouchard01:
Also updated readme

PR #160, opened 2026-02-10
Adding option to sample posterior in transformed (model) space

epaillas:
When the emulator is trained on a transformed representation of a summary statistic (e.g. log P(k)), but posterior inference is performed in the original, untransformed space, the implied noise model becomes inconsistent: the training loss encourages approximately Gaussian residuals in the transformed space, while the likelihood in linear space typically assumes additive Gaussian errors. This mismatch can lead to subtle biases (e.g. from exponentiation and Jensen’s inequality) and to an incorrect treatment of emulator uncertainty, which is intrinsically multiplicative in linear space.

While the above effects are expected to be small if the emulator uncertainty is subdominant with respect to the data uncertainty, we would like to be able to test this at the parameter inference level. This PR addresses the issue by allowing posterior sampling directly in the transformed space, ensuring that the likelihood, covariance, and emulator error model are defined consistently with the representation learned by the network.

### DIFF
--- a/acm/observables/combined.py
+++ b/acm/observables/combined.py
@@ -249,11 +249,16 @@ class CombinedObservable():
             # Transform each observable's samples, then concatenate
             cov_y_list = []
             for observable in self.observables:
-                # Get samples and transform them
-                observable._validate_output_transform()
-                cov_y_obs = observable.get_covariance_y(nofilters=False)
-                cov_y_transformed = observable.apply_output_transform(cov_y_obs)
-                cov_y_list.append(cov_y_transformed)
+                # Get samples and transform them if the observable has a transform
+                if hasattr(observable.model, 'transform_output') and observable.model.transform_output is not None:
+                    observable._validate_output_transform()
+                    cov_y_obs = observable.get_covariance_y(nofilters=False)
+                    cov_y_transformed = observable.apply_output_transform(cov_y_obs)
+                    cov_y_list.append(cov_y_transformed)
+                else:
+                    # No transform available, use regular samples
+                    cov_y_obs = observable.get_covariance_y(nofilters=False)
+                    cov_y_list.append(cov_y_obs)
             cov_y = np.concatenate(cov_y_list, axis=-1)
         else:
             cov_y = self.covariance_y
--- a/acm/observables/base.py
+++ b/acm/observables/base.py
@@ -608,13 +608,6 @@ class Observable():
         np.ndarray
             The combined data covariance matrix.
         """   
-        # if transform_output:
-        #     self._validate_output_transform()
-        #     # Get samples directly via method call (properties can't accept parameters)
-        #     cov_y = self.get_covariance_y(nofilters=False)
-        #     # Transform the samples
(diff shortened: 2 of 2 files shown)


acm/observables/combined.py
```diff
@@ -222,16 +245,28 @@ def get_covariance_matrix(
         np.ndarray
             The combined data covariance matrix.
         """   
-        cov_y = self.covariance_y
+        if transform_output:
+            # Transform each observable's samples, then concatenate
+            cov_y_list = []
+            for observable in self.observables:
+                # Get samples and transform them
+                observable._validate_output_transform()
```
SBouchard01: This will crash if one of the Observables has no `output_transform`

acm/observables/combined.py
```diff
@@ -222,16 +245,28 @@ def get_covariance_matrix(
...
+                cov_y_obs = observable.get_covariance_y(nofilters=False)
+                cov_y_transformed = observable.apply_output_transform(cov_y_obs)
+                cov_y_list.append(cov_y_transformed)
+            cov_y = np.concatenate(cov_y_list, axis=-1)
+        else:
+            cov_y = self.covariance_y
```
SBouchard01: This defeats a bit the goal of the class to handle the observables separately and concatenate the outputs. Here the transform_output argument only works if all models have a `transform_output` property defined

acm/observables/combined.py
```diff
@@ -257,16 +295,126 @@ def get_emulator_covariance_matrix(self, prefactor: float = 1, method: str = 'me
...
+            The block-diagonal transformed emulator covariance matrix.
+        """
+        return self.get_emulator_covariance_matrix(
+            prefactor=prefactor,
+            method=method,
+            diag=diag,
+            transform_output=True,
+            **kwargs
+        )
+
```
SBouchard01: Should these be defined in the `CombinedObservable` class ?

SBouchard01 (changes requested):
While the point of this PR is very interesting, overall I feel like the transformation stuff should be done at the model level (in `sunbird`), not in the Observable classes. I think this will complexify the base code foe the observables.

The transformed combined function should also be defined in a child class of BaseObservable & CombinedObservable, not in the main one as they are for specific use cases.

The main interesting point is the implementation of a transform function for the data, but we should define a child class of the Observable class that could handle this by overloading the required functions (and try to put that directly in `__getattr__` instead of all the call functions if that transform need to be applied at the output of the dataset elements ?). Adding a transform element in the output of `__getattr__` can be interesting, but the curent implementation has a lot of moving pieces that repeat themselves

acm/observables/emc/base.py
```diff
@@ -406,3 +433,133 @@ def compress_emulator_error(self, save_to: str = None):
...
+            If no output transform is available on the model.
+        """
+        return self.get_emulator_covariance_matrix(
+            prefactor=prefactor,
+            method=method,
+            diag=diag,
+            transform_output=True,
+            **kwargs
+        )
+
```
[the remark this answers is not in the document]
epaillas: Same here, are you suggesting we put them somewhere else?

PR #168, opened 2026-02-17
Add support to save clustering statistics in lsstypes (and other) formats

epaillas:
At the moment, the clustering estimator classes produce different kinds of outputs depending on the statistics (collection of correlation/power multipoles, coefficients, etc). Sometimes the data is saved as a simple 1D array (e.g. WST), sometimes as lists of pycorr objects (e.g. density split), etc, which can be confusing.

This PR begins to implement new `save` functions for the estimator classes, which can be used to save data as we currently are (for retrocompatibility), but more importantly, it allows us to save all statistics in a common `lsstypes` format (https://github.com/adematti/lsstypes), following the new standard for the clustering measurements produced by GQC. It also allows other formats (.nc or .zarr for xarray; only working for WST for now).

Output data format is given by the file suffix:
```
>>> wst.save('wst_coefficients.hdf5')  # lsstypes format
>>> wst.save('wst_coefficients.nc')     # xarray NetCDF
>>> wst.save('wst_coefficients.zarr')   # xarray Zarr
>>> wst.save('wst_coefficients.npy')    # numpy format
```

We can then read the data as e.g.

```
from lsstypes import read

data = read('ds_xiqg.h5')
xi = data.get(quantiles=4)
poles = xi.project(ells=[0, 2, 4])
```

### DIFF
--- a/acm/estimators/galaxy_clustering/base.py
+++ b/acm/estimators/galaxy_clustering/base.py
@@ -1,6 +1,12 @@
 import time
 import logging
+from typing import Optional
+from pathlib import Path
 
+import lsstypes as types
+import xarray as xr
+
+import numpy as np
 import numpy.typing as npt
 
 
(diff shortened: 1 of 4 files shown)


epaillas:
Ignore the "unit test for WST" commit, which is being reviewed in a separate PR.

PR #172, opened 2026-02-25
bug fix for cutsky n(z) and lightcone __init__

hbrincon:
This branch resolves a [missing factor of 3](https://github.com/epaillas/acm/blob/669a38cf5953e6df4422059f72f3210683de3563/acm/hod/cutsky.py#L746) in the calculation of the cutsky raw target density (due to the shape of `box_positions` being `(N, 3)`) and updates the lightcone __init__ function to [set up mpi](https://github.com/epaillas/acm/blob/669a38cf5953e6df4422059f72f3210683de3563/acm/hod/lightcone.py#L29) and thus stop the class from raising an error when it attempts to access inherited CutskyHOD methods.

### DIFF
--- a/acm/hod/cutsky.py
+++ b/acm/hod/cutsky.py
@@ -743,7 +743,7 @@ class CutskyHOD(BaseCutskyCatalog):
                 )
             # Each rank has a portion of box_positions after scatter
             # Need to gather total count for raw_nbar calculation
-            total_particles = mpy.csize(box_positions, mpicomm=self.mpicomm)
+            total_particles = mpy.csize(box_positions, mpicomm=self.mpicomm) // 3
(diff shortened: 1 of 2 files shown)


hbrincon:
Also relevant, I've noticed that the `cutsky.py` now makes use of fields, methods, and parameters named `csize` is various places. The n(z) bug was related to one of these instances. I'm unsure if the other instances may have produced similar bugs, but the cutsky output looks ok with the current fix.

Anyone who has used this branch for cutsky mocks may need to regenerate their mocks now that the bug is resolved.

epaillas:
Excellent. This looks in good shape so I'm going to approve it and re-generate the cutsky mocks. Thanks!

PR #180, opened 2026-03-05
Elg cutsky

hbrincon:
An initial setup for running ELG cutsky mocks. The ELG mocks will require further bug testing before they are ready to be implemented, so the option to run them is disabled for the time being.

More immediately, the branch includes bug fixes for the n(z) downsampling used for the lightcone and cutsky mocks. Previously, the hardcoded redshift binning used for the downsampling was too coarse, resulting in fluctuations in the downsampled n(z) away from the reference n(z) file. Now the binning is a user-specified parameter with a default value that provided accurate n(z) downsampling. @willlake and I contributed to this solution.

As part of the n(z) bugfix, I've commented out the `calculate_raw_nbar` method and its usage, as it's now been replaced by calculations on n(z) in spherical shells. I've left the commented out method present in case it is needed for future functionality.

### DIFF
--- a/.env
+++ b/.env
@@ -1,3 +1,0 @@
-#PYTHONPATH=/global/homes/j/jcolley/projet/acm
-PYTHONPATH=./
-PYTHONNOUSERSITE=0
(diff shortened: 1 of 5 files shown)


epaillas:
I'll leave the resolution of merge conflicts to you so that I don't break anything

PR #181, opened 2026-03-05
Hod

ntbfin00:
Added a function to check that the actual number density and satellite fraction of the catalogue matches the prediction from the HMF to some tolerance.

Added `save_distortions` argument to allow backwards compatibility with column format of current catalogues.

Predicted number density is now stored as class attribute and saved to file in the mock generation script.

### DIFF
--- a/acm/hod/box.py
+++ b/acm/hod/box.py
@@ -269,6 +269,9 @@ class BoxHOD:
             Whether to use the logarithm of sigma as the parameter for the HOD instead of sigma itself. 
             This is useful for sampling purposes, since sigma can vary over several orders of magnitude. 
             Default is True.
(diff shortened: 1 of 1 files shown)


acm/hod/box.py
```diff
@@ -207,6 +242,7 @@ def run(
         save_fn: str | Path | None = None,
         add_ap: bool = False,
         use_logsigma: bool = True,
+        save_distortions: bool = False,
```
SBouchard01: Can you add this parameter to the docstring too (also for `save_catalog`) ?

SBouchard01:
To make sure I understand correctly: the downsampling of the catalog is now done trough the `ic` parameter instead of a random subsampling, right ?

Issue #116, opened 2025-12-06
Observables aliases


SBouchard01:
An idea to make the Observables call easier : 
The main philosophy is to have very explicit Observables name, with a doctstring that explicits what this observable is used for.

But this can make very long class names to load (e.g. `GalaxyCorrelationFunctionMultipoles` and that's not the worst we have) 

Each of those Observables contains a unique `stat_name` value that is a shortened name that corresponds to the compressed file loaded and is re-used within the class.

I propose we create class aliases that match (when possible) this stat_name value. In the case of `GalaxyCorrelationFunctionMultipoles`, we have `stat_name='tpcf'`. We can the define the following alias : 

```python 
tpcf = GalaxyCorrelationFunctionMultipoles 
```

To make the namespace clearer, I'd recommend importing the aliases only in `__init__.py`, and the full classes names from the files themselves to prevent confusion ? 🤔

e.g. 
```python
from acm.observables.emc import tpcf # this is the alias of the GalaxyCorrelationFunctionMultipoles class
from acm.observables.emc.tpcf import GalaxyCorrelationFunctionMultipoles # This is the exact class name
from acm.observables.emc import GalaxyCorrelationFunctionMultipoles # This will crash
```

> [!NOTE]
> Ideally, we should prevent the files names and the aliases to have the same name to avoid confusion in the namespace 🤔 (e.g. `acm.observables.tpcf` should not include a `tpcf` alias - in this specific case we should change the name of the `tpcf.py` file to another name ?)

PR #190, opened 2026-03-25
[Feature] Observables aliases

SBouchard01:
Solves #116 for EMC and BGS observables

> [!NOTE]
> Removes the project observables from the package namespace, keeping only the aliases. This still does not solve the remark in #116 causing the namespace to have duplicates (e.g. `acm.observables.emc.tpcf` being a class and a module at the same time...)
> @luckyjim if you have suggestions on this, feel free to comment :)

### DIFF
--- a/scripts/emc/inference/inference_abacus.py
+++ b/scripts/emc/inference/inference_abacus.py
@@ -3,9 +3,9 @@ from sunbird.inference import priors as sunbird_priors
 from sunbird.cosmology.model_params import get_model_params
 from sunbird.inference.samples import Chain
 
-import acm.observables.emc as emc
 from acm.observables import CombinedObservable
 from acm.utils.covariance import get_covariance_correction
+from acm.utils.modules import get_class_from_module
 from acm import setup_logging
 
 from cosmoprimo import fiducial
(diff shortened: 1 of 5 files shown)


SBouchard01:
@epaillas I've updated the BGS files like yours and rebased the branche on dev. If all tests pass and your review is OK, , it should be all good to merge :)

scripts/emc/measurements/compress_files_aliases.py
```diff
@@ -0,0 +1,98 @@
...
+args = parser.parse_args()
+statistic = resolve_statistic_name(args.statistic)
+n_hod = args.n_hod
+add_covariance = args.add_covariance
+
+setup_logging()
+
+paths = lookup_registry_path("projects.yaml", "emc")
+
+observable = getattr(emc, statistic)
```
SBouchard01: @epaillas a small suggestion: I've made [this utils function](https://github.com/epaillas/acm/blob/84ca8dd33909f4dc9013c8af2054dd4adc76cd05/acm/utils/modules.py#L3) to not have to hardcode `emc` in the code if you want :)

e.g.
```python
parser.add_argument(
    '--module', 
    default='acm.observables.emc', 
    help='Module to load the Observable classes from'
)

[...]

observable = get_class_from_module(module, statistic)
```

SBouchard01: resolving as it is not really relevant in this case, we could use that if we want scripts regardless of project observables eventually

epaillas: Thanks! I added it in [d362dd4](https://github.com/epaillas/acm/pull/190/commits/d362dd4fb6b1bd926a2c0f7a5800c1f8d781fa1b)

PR #212, opened 2026-04-03
Rewrite fourier saved files

SBouchard01:
Since the write method of the lsstypes save tends to sometime cause corrupted files (0.01%), I try to mitigate that by writing to a temporary file and moving the final result as an atomic move. (by Arnaud's recommendation)

Not sure if that will solve the issue, but it's worth a try 🤷🏼


SBouchard01:
Not the right branch

PR #213, opened 2026-04-03
Rewrite fourier saved files

SBouchard01:
Since the write method of the lsstypes save tends to sometime cause corrupted files (0.01%), I try to mitigate that by writing to a temporary file and moving the final result as an atomic move. (by Arnaud's recommendation)

Not sure if that will solve the issue, but it's worth a try 🤷🏼

### DIFF
--- a/acm/estimators/galaxy_clustering/density_split.py
+++ b/acm/estimators/galaxy_clustering/density_split.py
@@ -125,7 +125,7 @@ class DensitySplit(BaseEstimator):
                 "Supported extensions are: .hdf5, .h5, .npy"
             )
 
-    def save_powers(self, powers, filename, attrs=None):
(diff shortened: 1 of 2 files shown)


epaillas (changes requested):
The atomic-write change looks good overall, but I agree with the existing inline comments on DensitySplit: the new tree.write(tmp_filename) + replace(path) paths should still be restricted to a single JAX process. PowerSpectrumMultipoles.compute_spectrum() already does this with jax.process_index() == 0, but save_correlations() and save_powers() now perform file I/O unconditionally. In multi-process runs that can race on the same temp file and final path, which undermines the corruption fix. Once those writes are single-writer guarded, this looks in good shape.

epaillas (approved):

epaillas:
Thanks for the changes!

PR #214, opened 2026-04-06
Toggle RSD for cutsky/lightcone mocks

hbrincon:
Removes hardcoded RSD application to cutsky/lightcone mocks and allows user to configure the desired behavior

### DIFF
--- a/acm/hod/cutsky.py
+++ b/acm/hod/cutsky.py
@@ -764,6 +764,7 @@ class CutskyHOD(BaseCutskyCatalog):
         release: str = 'Y1',
         target_nz_filename: str | None = None,
         custom_xyz_file: str | None = None,
+        apply_rsd: bool = True,
         nfw_draw_path: str = '/global/cfs/projectdirs/desi/users/arocher/nfw.npy'
     ) -> dict:
         """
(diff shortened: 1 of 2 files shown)


epaillas (changes requested):
Thanks for making RSD configurable here. I found one blocking regression: when apply_rsd=False, box_to_cutsky() does not populate RSDPosition, but both CutskyHOD.sample_hod() and LightconeHOD.sample_hod() still unconditionally copy every key from self.keys_cutsky / self.keys_lightcone, which both include RSDPosition. That means disabling RSD will fail while building the catalog instead of returning a real-space mock. Could you either keep RSDPosition populated in the no-RSD case (for example as Position) or make the output schema/append loop conditional on apply_rsd?

hbrincon:
I think it would make the most sense to make the output schema/append loop conditional on apply_rsd. I've made it so that the keys_lightcone/keys_cutsky lists are updated each time setup_hod is called. @epaillas let me know if the changes look good with you

epaillas (approved):

epaillas:
Looks good, thanks! Feel free to merge after tests pass

Issue #201, opened 2026-03-31
[Feature] Add formatter CI


SBouchard01:
With #44, we need to add a workflow to correctly format the new code in PRs to `main` and `dev`

luckyjim:
I was able to add a GitHub "repository secret" without any special permission :
https://github.com/epaillas/acm/settings/secrets/actions
This allows you to commit/push during CI.

After add ruff  in workflow in test-ci branch , the CI correctly format code and run pytest:
 https://github.com/epaillas/acm/actions/runs/24183529161

luckyjim:
Creation of a specific branch for pull request

luckyjim:
The checkout of package when it's a pull request trigger the workflow is different of push trigger.
https://docs.github.com/fr/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request

<img width="918" height="163" alt="Image" src="https://github.com/user-attachments/assets/056b6932-b409-4de9-9381-4bd752fce68b" />


workflow checkout by default a branch :

* pull/PULL_REQUEST_NUMBER/merge

and not the the source branch, so some modification is necessary to do that, like

 
  ```
 - name: Checkout
        uses: actions/checkout@v4
        with:
          ref: ${{ github.event.pull_request.head.ref }}
          persist-credentials: false

```
persist-credentials: false

 is necessary to used the repository secret else you can't commit push modification.

The second problem is that formatting creates a new, synchronized workflow.

To avoid running pytest (~ 15 or 20 mn) twice, the first workflow performs only the formatting, and the second only runs pytest.

The case where no formatting is applied is handled.

PR #218, opened 2026-04-09
add format in CI

luckyjim:
See #201 

Since this pull request will trigger the CI/CD pipeline and contains the reformatting of the code that will be executed the first time, many files will be modified.

Before accepting the PR  it is necessary to merge it with the latest development version to avoid conflicts. I can do it if you want.


luckyjim:
There's a difference in behavior between the pull request trigger used here and the push-on trigger on this branch used for validation 

https://github.com/epaillas/acm/actions/runs/24183529161

that I don't understand.

luckyjim:
Pull request used to develop the GitHub action format workflow.
it's OK now, use of new clean branch with  GitHub action solution

PR #222, opened 2026-04-14
Update test ci to prevent concurrency and split jobs

### LINKED ISSUE #221: [Feature] Rework format CI
1. Split format job in 2: check which jobs to run then format
2.  Add ruff install action
3. Handle job concurency to prevent test from running several times
4. Use ssh key instead of token to allow pushes on main and dev
5. Split format step in more sub-steps (checkout/git setup/ruff install/ruff format/commit/push)

SBouchard01:
pytest now only runs when no formatting is applied thanks to concurency canceling in-progress workflows


SBouchard01:
@luckyjim this needs to be tested on a fork with the `secrets.PUSH_KEY` setup (public ssh key in deploy keys, and private part in github secrets)

Let me know if this work as intended ! :)

SBouchard01:
Tested on a fork:
- https://github.com/SBouchard01/acm/pull/2 (branch formatting --> dev)
- https://github.com/SBouchard01/acm/pull/4 (dev w/ unformatted hotfix commit --> main)

You can check the last commits of each PR, it works using action concurrency to cancel previous jobs:
1. First workflow triggers with reformat & tests
2. If reformat commit happens, a new workflow is triggered that cancels the 1. Since no reformat commit happens, the tests run.
3. If another commit is added to the PR, the previous workflow (1. or 2.) is cancelled and starts again from 1.

PR #224, opened 2026-04-16
Feature/logger

### LINKED ISSUE #198: [Feature] Logger : display by default the module and the line associated with the log message
Actually the logger returns message like this:

```
[000121.00]  07-01 13:45  CutskyHOD                    INFO     Applying RSD.
[000123.61]  07-01 13:45  CutskyHOD                    INFO     Converting to sky positions.
[000128.73]  07-01 13:45  CutskyHOD                    INFO     Applying radial mask.
[000130.81]  07-01 13:45  CutskyHOD                    INFO     Applying footprint mask.
[000140.73]  07-01 13:45  Footprint                    INFO     Load DR9 footprint with nside=256, mask_lmc=False,
```

without the line number associated with the log command and only the class name and not the module path in the library, especially considering that there are multiple classes with the same name, it's ambiguous.

Image

-- 

These two pieces of information allow you to quickly find the line associated with the log, which is not possible with current messages.

This allows for a more systematic way of defining the logger associated with the module and simplifies the definition of classes that no longer handle this functionality.

This is also the approach suggested in the Python documentation for using loggers.

```
A good convention to use when naming loggers is to use a module-level logger, in each module which uses logging, named as follows:

logger = logging.getLogger(__name__)
```

https://docs.python.org/3.8/howto/logging.html#advanced-logging-tutorial

Here's what the messages would look like after modification.

```
[000121.00]  07-01 13:45  [acm.hod.box.cutsky  20]                    INFO     Applying RSD.
[000123.61]  07-01 13:45  [acm.hod.box.cutsky  100]                  INFO     Converting to sky positions.
[000128.73]  07-01 13:45  [acm.hod.box.cutsky  120]                  INFO     Applying radial mask.
```

The current interface utils.logging.py would not be changed, but all modules will have

```
from logging import getLogger

logger = getLogger(__name__)
```

If you agree, I can make the necessary changes to the project.

luckyjim:
from #198 

Documentation is in header of module logging.py

https://github.com/epaillas/acm/blob/feature/logger/acm/utils/logging.py

script logger is only used by 

https://github.com/epaillas/acm/blob/feature/logger/tests/test_minkowski.py


acm/utils/logging.py
```diff
@@ -73,13 +135,66 @@ def mkdir(dirname):
...
+
+    return
+    ------
+    string like NAME_PKG_GIT.xx.yy.zz of script that call this function
+    """
+    l_sep = osp.sep
+    r_str = l_sep + NAME_PKG_GIT + l_sep
+    p_grand = pfile.find(r_str)
+    if p_grand > 0:
+        # -3 for size of ".py"
```
SBouchard01: This means the logger only returns the correct name if the extension length is 3.
Maybe we could make this slightly more customisable (to ensure it causes no weird display if called from another filetype e.g. `.ipynb`) by either passing the extension length as an extra arg defaulting to 3, or directly dropping the extension with `pathlib.Path().with_suffix('')` which is OS-agnostic ?

acm/utils/logging.py
```diff
@@ -73,13 +135,66 @@ def mkdir(dirname):
...
+        g_str = pfile[p_grand + 1 : -3].replace(l_sep, ".")
+    else:
+        # out package git
+        # -3 for size of ".py"
+        print("out package git")
```
SBouchard01: Is that `print` left here intentionally ?

acm/utils/logging.py
```diff
@@ -73,13 +135,66 @@ def mkdir(dirname):
...
+    Parameters
+    ----------
+    pfile: path of the file,
+
+    return
+    ------
+    string like NAME_PKG_GIT.xx.yy.zz of script that call this function
+    """
+    l_sep = osp.sep
+    r_str = l_sep + NAME_PKG_GIT + l_sep
```
SBouchard01: `NAME_PKG_GIT` being hardcoded in the file means we could not import this from other eventual packages. Can we set it as an optional function argument (defaulting to `"acm"`)?

acm/utils/logging.py
```diff
@@ -73,13 +135,66 @@ def mkdir(dirname):
...
+    Note: this function should be called before setup_logging()
+
+    Parameters
+    ----------
+    pfile: path of the file, so always call with __file__ value
+
+    """
+    # print(pfile)
+    str_logger = _get_logger_path(pfile)
+    # print("logger name: %s" % str_logger)
```
SBouchard01: [nitpick] [cleanup]: remove if not needed

acm/utils/logging.py
```diff
@@ -73,13 +135,66 @@ def mkdir(dirname):
...
+    so that we can have different loggers for different scripts.
+
+    Note: this function should be called before setup_logging()
+
+    Parameters
+    ----------
+    pfile: path of the file, so always call with __file__ value
+
+    """
+    # print(pfile)
```
SBouchard01: [nitpick] [cleanup]: remove if not needed

SBouchard01 (commented):
Thank you very much ! :) I added some very minor suggestions

Also, this seems to branch from `feature/formatter-ci` instead of `dev` (this might cause merge issues later 🤔 ) 
EDIT: If rebased on the branch after the workflow trigger is fixed, it should be good :)

luckyjim:
@copilot resolve the merge conflicts in this pull request

luckyjim:
ok I will refactoring _get_logger_path() function and also solve conflicts

luckyjim:
for script out of acm package writes only the name of script without path in line logger. The absolute path of script is write at the begging of the script, example:

```
ogin27:~>cp projet/acm/tests/test_minkowski.py projet/ 
(base) NERCS: ~
login27:~>python projet/test_minkowski.py 
[000000.00]  04-17 03:49  acm.utils.logging:111 INFO Start logger at 2026-04-17 03:49:54
[000000.00]  04-17 03:49  acm.utils.logging:207 INFO 
[000000.00]  04-17 03:49  acm.utils.logging:207 INFO Full name script: /global/u2/j/jcolley/projet/test_minkowski.py
[000000.00]  04-17 03:49  acm.utils.logging:207 INFO 
[000000.64]  04-17 03:49  test_minkowski.py:138 INFO Loaded HOD catalog: /pscratch/sd/n/ntbfin/emulator/hods/z0.5/yuan23_prior/c000_ph000/seed0/hod000.fits
[000000.64]  04-17 03:49  test_minkowski.py:139 INFO *****************
[000000.64]  04-17 03:49  test_minkowski.py:139 INFO Number of galaxies: 4000000
[000000.64]  04-17 03:49  test_minkowski.py:139 INFO *****************
[000000.64]  04-17 03:49  test_minkowski.py:140 INFO Box size: [2000. 2000. 2000.]
[000000.64]  04-17 03:49  acm.estimators.galaxy_clustering.base:20 INFO Initializing MinkowskiFunctionals.

```

luckyjim:
I add this when set_logging is called
[000000.00]  04-17 03:49  acm.utils.logging:111 INFO Start logger at 2026-04-17 03:49:54

may be remove the date at each line of logger ?

SBouchard01:
> may be remove the date at each line of logger ?

I think it can be useful if we have a job spanning on several days or missing parts of the logging (namely the beginning) , so I'd leave it there for now

acm/utils/logging.py
```diff
@@ -100,3 +153,68 @@ def suppress_logging(enabled=True, highest_level=logging.CRITICAL):
...
+    if pkg_name in parts:
+        idx = parts.index(pkg_name)
+        module_parts = parts[idx:]  # keep only the parts after the package name
+    elif parts[0] == pfile.drive + pfile.root:
+        module_parts = parts[1:]  # remove the drive and root if it's an absolute path
+    else:
+        module_parts = parts
+
+    logger_name = ".".join(module_parts)
+    return logger_name
```
SBouchard01, line 191: Duplicate of the next function

acm/utils/logging.py
```diff
@@ -100,3 +153,68 @@ def suppress_logging(enabled=True, highest_level=logging.CRITICAL):
...
+
+    Parameters
+    ----------
+    pfile: path of the file,
+
+    return
+    ------
+    string like NAME_PKG_GIT.xx.yy.zz of script that call this function or just name script
+    """
+    logger.info(f"\nFull name script: {pfile}\n")
```
SBouchard01, line 211: Could be on debug level ? I guess it's not really needed in INFO 🤔

acm/utils/logging.py
```diff
@@ -60,6 +110,9 @@ def format(self, record):
     handler.setFormatter(fmt)
     logging.basicConfig(level=level, handlers=[handler], **kwargs)
     sys.excepthook = exception_handler
+    logger.info(
+        f"Start logger at {datetime.datetime.now().strftime('%Y-%m-%d %H:%M:%S')}"
+    )
```
SBouchard01, line 115: Could be on debug level ? I guess it's not really needed in INFO 🤔

SBouchard01 (commented):
1 duplicate and maybe 2 INFO/DEBUG levels to discuss

All good otherwise :)

luckyjim:
I remove duplicate code.
A log file traditionally begins with the date and time, as well as the full path of the script executed.

If you keep the date in each log message, the first line indicating the date and time is unnecessary; you can delete it. As for the script name, the user can always add it themselves; you can also delete it or change the log level as you wish.

SBouchard01:
I think these are useful informations, but can be duplicate and/or misused (e.g. if nested scripts call `_get_logger_path`). 

Also, the INFO level will have the logger initialization message at each first package call which can make the logs noisy. 

I'll keep them but move them at DEBUG level to avoid INFO noise with packages calls. 

If you are OK with that I can push the change and merge the branch after ? 👍

PR #226, opened 2026-04-16
Update cache key in CI workflow to use pyproject.toml

SBouchard01:
Changed the reference file as dependencies are listed in the pyproject files (so that its hash is updated when we update the dependencies)

Also remove caching in format workflow to prevent caching pip env without all dependencies (also the ruff instal is ~4s, I'm not sure the cache will be faster)

### DIFF
--- a/.github/workflows/ci-format-tests.yml
+++ b/.github/workflows/ci-format-tests.yml
@@ -30,14 +30,6 @@ jobs:
         with:
           python-version: "3.13"
 
-      - name: Cache installation
-        uses: actions/cache@v3
-        with:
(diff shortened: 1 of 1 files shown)


SBouchard01:
Adding conditional test CI run in [9f0757d](https://github.com/epaillas/acm/pull/226/commits/9f0757d0e263875490ee5156524918ba8f8fe638) to skip tests if reformat is pushed to repo (as a new workflow will be triggered anyways)

SBouchard01:
Minor changes and tests pass, merging

Issue #196, opened 2026-03-25
Refactor acm.hod code


SBouchard01:
The current implementation of the classes generating galaxy catalogs in `acm.hod` makes it hard to test: 
- Some methods use files with hardcoded paths
- No common implementation between the 3 geometries 
- Only one tracer available per class call (#124)

I propose an to implement a class that loads the dark matter catalogs and calls a HOD sampler to create galaxy catalogs, handled by another class.
All geometries would then inherit from those base classes. Those classes would have dedicated methods for each operation, to allow an easier access to each step of the run, and easier testing and debugging.

I created a feature branch with #195 to develop this. It would also solve #124, #54 and #178 

> [!NOTE]
> Eventually, I'd also like to depend less on `AbacusHOD` by being able to set a backend to load and populate the dark matter catalogs - just like we did for the estimator classes.

SBouchard01:
Most of this work has been done in acm.catalogs (from #195 ), we still need to port the `acm.hod` functionalities regarding cutsky and lightcone - this development is currently lead by @hbrincon

Issue #184, opened 2026-03-19
[Feature] Remove code duplication in the compress_data() method in the derived BaseObservable classes


luckyjim:
@SBouchard01  noticed code duplication with the compress_data() method. 

The pylint static code analyzer confirms this.

```
$ pylint --disable=all --enable=duplicate-code acm/acm > res_dupli.txt
$ wc -l res_dupli.txt
4200 
```
so around 4200 lines duplicated (but included docstring ) almost 

```
$ grep "def " res_dupli.txt | wc
     66     367    3589
```

in  compress_data() and plot_xx()  method

```
$ grep "def compress_data" res_dupli.txt | wc
     19      38     437

$ grep "def plot" res_dupli.txt | wc
     42     286    2750
```

I'm checking if there's a refactoring solution that can limit this ...

SBouchard01:
Adding some reflexions here: 
- If we standardize estimator output to lsstypes, we could make this a standard method
- For reading purpose, I would also add the x parameters values in the output files metadata (see #179), with an option to provide for it externally. That would ensure the x-y mapping is consistent whatever the files we explore
- I can find a way to loop over any set of parameters if a glob search pattern is provided as an input (something like `'c{cosmo_idx}_ph{phase_idx}/seed{seed}/hod{hod_idx}'` - still need to figure out the filename formats to allow lsstypes.mean)

SBouchard01:
Added compression utils in #245 

When needed, we can remove all the compression & plot methods from the observable classes

PR #227, opened 2026-04-17
[Feature] Add linter/typecheck CI

### LINKED ISSUE #199: Status and correction of type errors
[ty](https://docs.astral.sh/ty/) the type checker used by script acm_type.sh indicates around:
* 400 errors
* 50 warning

in acm source.

Most of them are due to the fact that : 
* the annotation is not up to date 
 
```
error[invalid-return-type]: Return type does not match returned value
   --> acm/observables/emc/bispectrum.py:228:16
    |
226 |             np.save(save_fn, dataset_to_dict(cout))
227 |             logger.info(f'Saving compressed data to {save_fn}')
228 |         return cout
    |                ^^^^ expected `dict[Unknown, Unknown]`, found `Dataset`
229 |     
230 |     @set_plot_style
    |
   ::: acm/observables/emc/bispectrum.py:118:10
    |
116 |         seed: int = 0,
117 |         test_filters: dict = None,
118 |     ) -> dict:

```

* false positive

```
error[unresolved-import]: Module `scipy.spatial` has no member `cKDTree`
 --> acm/estimators/galaxy_clustering/multiplets.py:5:27
  |
5 | from scipy.spatial import cKDTree 
  |                           ^^^^^^^
6 | from scipy.sparse import csr_matrix
7 | from scipy.sparse.csgraph import connected_components
  |
info: rule `unresolved-import` is enabled by default
```
Can be solve by : 

`
from scipy.spatial import cKDTree # ty: ignore[unresolved-import]`

Therefore, currently, type checking cannot identify code regressions such as interface changes that haven't propagated throughout the code.

Thus, I encourage developers to update the annotations of the code they modify.

I start this work in branche [feature/observable_quality](https://github.com/epaillas/acm/tree/feature/observable_quality) for https://github.com/epaillas/acm/blob/feature/observable_quality/acm/observables/base.py

### LINKED ISSUE #202: [Feature] Add linting CI
With #44 we also need a linter workflow to _at least_ outline the mistakes in PRs code quality (and _at best_ automatically fix some of those)

SBouchard01:
This typecheck should ignore missing imports (as I don't load the env on purpose to avoid ~8m of compilation time), and returns anotations

> [!WARNING]
> This branch is still experimental, I'm testing what's best between Commit annotations and markdown PR comment


SBouchard01:
> [!TIP]
> The annotations will appear in the logs for now until we fix those.
> Once fixed, annotations for commited code in the PR will appear in the "Files changed" tab (see [b9a4e30](https://github.com/epaillas/acm/pull/227/commits/b9a4e30cfd80c29699ea82ff23be44972db3c6c4))

SBouchard01:
> [!CAUTION]
> To allow this branch to merge, we will need to fix the **317** type errors raised by `ty`.

SBouchard01:
The current implementation ignores on purpose the following files for type checking:
- `acm/hod/*` (as #196 will rework it it's not worth trying to fix the current implementation)

I've also set the invalid-method-override to warn only as we have several override in the Observable classes (this one is on me) that don't match their parent signature and I don't want to rewrite all of those (especially since I plan to remove all of that with #184 ), but might cause some issues eventually - we should keep an eye on it (see: https://stackoverflow.com/a/14770951 and https://en.wikipedia.org/wiki/Liskov_substitution_principle)

SBouchard01:
Update: I applied some lazy fixes to the `acm.hod` module, pending the rework so that we can check the linting passes there in the future. I'll probably ignore the many Ruff errors there because the rework will be faster than fixing everything in the current architecture (as it is the source of most flags raised).

> [!NOTE]
> There are still some `invalid-method-override` warnings in the `acm.hod` and `acm.observable` classes, but these should be removed with time as #184 and #196 will be worked on

@epaillas I think this PR is now ready to be reviewed, it has been rebased on `dev`.

epaillas:
I had identified an issue with the pyrecon backend, but I think copilot also flagged it and I see you've taken care of it:

PyreconBackend constructor is broken for preloaded catalogs on the PR branch. In acm/estimators/galaxy_clustering/backends/pyrecon.py:117-129, _assign_data() / _assign_randoms() run before the counters are initialized, and then size_data / _size_randoms are reset to 0 immediately afterward.

Codex also highlighted a few other things. Could you check the following:

  1. The new from typing import override imports make several observable modules unimportable on Python 3.10/3.11, even though pyproject.toml:21 now declares requires-python = ">=3.10". The bad imports are at acm/observables/bgs/density_split_correlation.py:3, acm/observables/emc/density_split_module.py:3, and acm/observables/emc/versus_module.py:3. In this environment on Python 3.11, from typing import override raises ImportError. CI won’t catch it because the workflow only runs Python 3.13.
  
  2. try_eval() no longer preserves ordinary string values. In acm/utils/parser.py:192-195, the catch was narrowed from bare except to except TypeError, but eval("foo") and many CLI-style values fail with NameError or SyntaxError, not TypeError. I verified that try_eval("foo") now raises NameError on the PR branch. Since args.paths, select_filters, and slice_filters all flow through this helper later in the file, unquoted string arguments that used to work now abort parsing.

SBouchard01:
@epaillas It should be fixed now. I plan to drop `acm.utils.parser` altogether, it was used in an old BGS script but with the implementation of the Observables, I don't think it's relevant anymore (unless you have an use for it, I think the use of aliases or just the base Observable class are more powerful...)

PR #228, opened 2026-04-17
Stabilize ty baseline for observables and dynamic modules

epaillas:
## Summary
- configure ty to analyze the project as Python 3.10 and add targeted overrides for dynamic modules that are not ready for strict checking
- tighten shared observable and utility typing so the remaining checked surface passes cleanly
- make xarray serialization helpers accept DataArray inputs used by the observable save paths

## Verification
- ty check acm
- python3 -m py_compile acm/observables/base.py acm/observables/bgs/base.py acm/observables/combined.py acm/observables/emc/base.py acm/utils/abacus.py acm/utils/parser.py acm/utils/plotting.py acm/utils/xarray.py

This is intended to unblock the ty job added in PR 227 by making the current baseline actionable.

### DIFF
--- a/acm/observables/base.py
+++ b/acm/observables/base.py
@@ -153,8 +153,8 @@ class Observable:
             )
 
         # Set the filters
-        self.select_filters = select_filters
-        self.slice_filters = slice_filters
+        self.select_filters = select_filters if select_filters is not None else {}
+        self.slice_filters = slice_filters if slice_filters is not None else {}
         self.select_indices = select_indices
         self.select_indices_on = (
             select_indices_on if select_indices_on else []
--- a/acm/observables/base.py
+++ b/acm/observables/base.py
@@ -936,15 +936,6 @@ class Observable:
 
         return cov
 
-    @overload
-    def get_save_handle(self, save_dir: None = None) -> str: ...
-
-    @overload
-    def get_save_handle(self, save_dir: str) -> str: ...
-
-    @overload
-    def get_save_handle(self, save_dir: Path) -> Path: ...
-
     def get_save_handle(
         self, save_dir: str | Path | None = None
     ) -> str | Path:
--- a/acm/observables/base.py
+++ b/acm/observables/base.py
@@ -2,6 +2,7 @@ import logging
 import pickle
 from copy import copy, deepcopy
 from pathlib import Path
+from typing import cast, overload
 
 import matplotlib.pyplot as plt
 import numpy as np
@@ -391,12 +392,26 @@ class Observable:
             raise TypeError("Expected a DataArray after filtering and flattening.")
         return dataarray
 
+    @overload
     @staticmethod
     def stack_on_attribute(
         attribute: str | dict, dataarray: xarray.DataArray, **kwargs
-    ) -> xarray.DataArray:
+    ) -> xarray.DataArray: ...
+
+    @overload
+    @staticmethod
+    def stack_on_attribute(
+        attribute: str | dict, dataarray: xarray.Dataset, **kwargs
+    ) -> xarray.Dataset: ...
+
+    @staticmethod
+    def stack_on_attribute(
+        attribute: str | dict,
+        dataarray: xarray.DataArray | xarray.Dataset,
+        **kwargs,
+    ) -> xarray.DataArray | xarray.Dataset:
         """
-        Stacks a DataArray on the dimensions given.
+        Stack an xarray object on the dimensions given.
 
         Parameters
         ----------
@@ -404,15 +419,15 @@ class Observable:
             The dimension(s) to stack on.
             If a string, will be read from the DataArray attributes.
             Will be used as the dim to stack on (see xarray.DataArray.stack)
-        dataarray : xarray.DataArray
-            The DataArray to stack the dimensions on.
+        dataarray : xarray.DataArray | xarray.Dataset
+            The xarray object to stack the dimensions on.
         **kwargs
             Additional keyword arguments to pass to the stack method.
 
         Returns
         -------
-        xarray.DataArray
-            The stacked DataArray
+        xarray.DataArray | xarray.Dataset
+            The stacked xarray object.
         """
         if isinstance(attribute, str):
             if not attribute in dataarray.attrs or attribute in dataarray.dims:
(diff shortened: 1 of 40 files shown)


pyproject.toml
```diff
@@ -118,6 +118,31 @@ acm = [
...
+
+[[tool.ty.overrides]]
+include = [
+    "acm/hod/*.py",
+    "acm/estimators/galaxy_clustering/**/*.py",
+    "acm/lensing/abacuslensing.py",
+    "acm/observables/emc/*_module.py",
+    "acm/observables/bgs/tpcf_module.py",
+    "acm/observables/bgs/density_split_correlation.py",
+]
```
SBouchard01, line 141: I would avoid ignoring package files in the type checking

pyproject.toml
```diff
@@ -118,6 +118,31 @@ acm = [
     "**/*.yaml",
 ]
 
+[tool.ty.environment]
+python-version = "3.10"
```
SBouchard01: I'd avoid fixing the Python version here - ty runs on the current version which would depend on whatever we run on NERSC btw (currently 1.12 ?)

acm/utils/xarray.py
```diff
@@ -1,7 +1,17 @@
...
 import xarray as xr
 
 
-def dataset_to_dict(dataset: xr.Dataset) -> dict:
+def _ensure_dataset(dataset: xr.Dataset | xr.DataArray) -> xr.Dataset:
+    """Return a Dataset view for serialization helpers."""
+    if isinstance(dataset, xr.DataArray):
+        name = dataset.name or "data"
+        return dataset.to_dataset(name=name)
+    return dataset
```
SBouchard01: This might lead to inconsistent behavior, I need to ensure I don't use DataArray properties somewhere 
--> Needs toe be checked before approval

acm/observables/combined.py
```diff
@@ -302,7 +302,7 @@ def get_save_handle(self, save_dir: str | Path = None) -> str | Path:
         statistic_handles = [
             observable.get_save_handle() for observable in self.observables
         ]
-        statistic_handle = "+".join(statistic_handles)
+        statistic_handle = "+".join(str(handle) for handle in statistic_handles)
```
SBouchard01: `statistic_handles` should only contain `str` element --> this is a typing issue, not sure we need a loop here

acm/observables/base.py
```diff
@@ -511,7 +530,7 @@ def apply_indices_selection(self, dataarray: xarray.DataArray) -> xarray.DataArr
         xarray.DataArray
             The DataArray with the selected indices.
         """
-        if self.select_indices is None:
+        if self.select_indices is None or isinstance(dataarray, xarray.Dataset):
```
SBouchard01: DataArray allows indice selection, this will cause issues

acm/observables/base.py
```diff
@@ -484,6 +499,8 @@ def flatten_output(
         xarray.DataArray
             The flattened DataArray.
         """
+        if isinstance(dataarray, xarray.Dataset) or flat_output_dims is None:
```
SBouchard01: DataArray works too, why filter here ?

acm/observables/emc/base.py
```diff
@@ -429,6 +433,7 @@ def compress_x(
 
         x = np.concatenate(x)
         x_names = x_cosmo_names + x_hod_names
+        assert n_hod is not None
```
SBouchard01: assertions in methods is not recommended, raise an error instead (though n_hod should not be allowed to be None ?)

acm/observables/emc/base.py
```diff
@@ -162,9 +163,10 @@ def get_emulator_error(self) -> xarray.DataArray | np.ndarray:
         )  # Unfiltered covariance array !
 
         # Flatten on 2D for indexing
-        emulator_covariance_y = self.flatten_output(
-            emulator_covariance_y, flat_output_dims=2
-        )
+        if isinstance(emulator_covariance_y, xarray.DataArray):
```
SBouchard01: Why the filter in the instance here ?

SBouchard01 (changes requested):
Branch is not on the correct branch and has implementations on instances that I need to check closer

epaillas:
Codex here. I addressed the `DataArray` / `Dataset` concern in the latest commit by switching the helper typing to overloads rather than relying on runtime `isinstance` gates.

The intent is that `flatten_output` and `apply_indices_selection` preserve whatever xarray type they receive, and the overloads make that explicit to `ty` as "same type in, same type out". That let me remove the `DataArray`-only flattening guard in the emulator-error path, so we are no longer skipping flattening for one xarray type just to satisfy the checker.

So the overloads here are only for type precision; they are not meant to change the behavior contract of those helpers beyond removing that accidental split.

epaillas:
Codex here. I pushed a follow-up commit to address the current review notes.

For the `pyproject.toml` comments, I removed the `tool.ty` Python-version pin and instead moved that signal to project metadata by setting `requires-python = ">=3.10"`, since the codebase already uses `X | Y` annotations broadly. I also narrowed the `all = "ignore"` override to explicit non-package files, so the package `__init__.py` files are no longer excluded from type checking.

For `acm/utils/xarray.py`, I reverted `dataset_to_dict()` back to a Dataset-only helper rather than broadening its behavior. The two places that were serializing a `DataArray` covariance now convert to a dataset explicitly at the call site, which felt closer to the original contract.

For `acm/observables/combined.py`, I addressed the typing issue by adding overloads to `get_save_handle()`, so `statistic_handles` is typed as `list[str]` when no save directory is passed and the join can stay direct.

For the `assert` comment in `acm/observables/emc/base.py`, I replaced that guard with an explicit `ValueError` if `cosmos` is empty. I made the same change in the analogous BGS path for consistency.

Codex verification after the change: `ty check acm` passes.

epaillas:
Codex here. I pushed a follow-up pass to remove the remaining `ty` failures after dropping the broad package ignore.

This update fixes the outstanding issues in code across the backend/estimator/HOD/observable stack: tighter optional handling and return paths, dataset/dataarray consistency, a few backend-specific guards, and small plotting/attribute cleanups in the void and lensing modules.

`ty check acm` now passes on `feature/fix-ty-errors`. The only config-level exception I kept is a narrow override for `acm/estimators/galaxy_clustering/knn.py`, because `ty` still does not understand SciPy `cKDTree` and numba `prange` stubs there, and forcing that one through in code would have added noise without changing behavior.

SBouchard01:
Iterating on this in #232 locally: I solved the last issues, ignoring some folders on purpose. Because of the merge issues, I propose closing this PR/branch and merging PR#232 if @epaillas validates it 👍🏼

PR #232, opened 2026-04-18
Fix linter CI errors for first merge

### LINKED ISSUE #199: Status and correction of type errors

SBouchard01:
from #228 with some changes

### Notes
1. I ignored all errors of the hod module as there were a lot of flags and we'll rework that soon anyways
2. Observables overriding the compression methods don't match the signature of the original function. This has been ignored on purpose (solving #184 among other things might solve this eventually)
3. In theory, some the galaxy clustering estimators backends could be reworked to share the same attributes and method names for consistency, but this has been partly solved here

### DIFF
--- a/.github/workflows/ci-format-lint-test.yml
+++ b/.github/workflows/ci-format-lint-test.yml
@@ -79,7 +79,9 @@ jobs:
           python-version: "3.13"
 
       - name: Installation package with editable mode
+        id: python-setup
         run: |
+          echo "Python version: $(python --version)">> $GITHUB_OUTPUT
           python -m pip install --upgrade pip
           pip install ty
 
@@ -87,8 +89,8 @@ jobs:
         # First run is to get the number of errors without failing
         # Second run is to get the output in GitHub format for annotations
         run: |
-          ty check acm --quiet --exit-zero
-          ty check acm --output-format github
+          ty check acm --quiet --exit-zero --python-version ${{ steps.python-setup.outputs.python-version }}
+          ty check acm --python-version ${{ steps.python-setup.outputs.python-version }} --output-format github 
           echo "Exit code: $?"
 
   job_pytest:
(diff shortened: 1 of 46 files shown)


epaillas:
I'm getting the errors you reported, but other than that this looks fine by me, so feel free to merge, and then we can close #228

SBouchard01:
All fixed, one or two errors may persist, I'll solve those on the lint-ci branch if they do.

PR #234, opened 2026-04-18
Read EMC density-split covariance from lsstype's h5 files

epaillas:
This pull request updates the `compress_covariance` function in `density_split_module.py` to update how measurement data is loaded and processed. We switch from `.npy` to `.h5` files, using the `BaseEstimator.read` method for data loading, and refining how simulation phase indices are extracted and stored so that the data array preserves the true abacus_small phase.

I would like to slowly move towards having everything stored in lsstypes format (for the supported statistics). Main concern is I/O. Reading thousands of files and applying the lsstype filtering on them (scale cuts and rebinning) seems to be a bottleneck. The compression scripts can take several minutes to run on a single of these statistics.

### DIFF
--- a/acm/observables/emc/density_split_module.py
+++ b/acm/observables/emc/density_split_module.py
@@ -4,6 +4,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 import xarray
 
+from acm.estimators.galaxy_clustering.base import BaseEstimator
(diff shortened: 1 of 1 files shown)


SBouchard01:
Yeah that's one of the issues I have currently with `lsstypes`, the I/O time is very noticeable for DensitySplit Observables :/

PR #241, opened 2026-04-20
Features/ruff rules

### LINKED ISSUE #233: [Feature] Define and fix Ruff checks rules
The curent implementation on `dev` only checks rules "I" (`isort`) with an auto fix, and "D" and "'PL" for auto formatting. We could consider adding several rules ("F" for Pyflake, E for PyCodeStyle, etc.)

There are a lot of rules, and some of them are more restrictive than other, we need to discuss how much we want to constrain the ruleset.

Some sources: 
- https://docs.astral.sh/ruff/rules/
- https://www.reddit.com/r/Python/comments/1kttfst/ruff_users_what_rules_are_using_and_what_are_you/
- https://wbarillon.medium.com/mastering-python-linting-keep-control-over-your-code-7aa371d31c36

SBouchard01:
Adding rules discussed in #233 and fixing all 892 errors raised from that ruleset

> [!WARNING]
> These fixes are done by hand an in theory should not break anything. However, some changes were more complex than others, and might lead to unstabilities if the methods are not used correctly. If this is the case, let us know.

> [!NOTE]
> From now on, the CI imposes these rules to be applied to merge. This allows for a fixed code quality, but imposes some rigor on typing and casting to ensure no bugs are raised from those issues.


SBouchard01:
Notes on fixing issues:
- MST is using dict keys with % symbol (e.g. `mst%ipt`) which is not recommended as it is an interpreted string formatting symbol. Ignoring rule UP031 on those lines for now

acm/estimators/galaxy_clustering/base.py
```diff
@@ -1,43 +1,39 @@
...
 import time
 from pathlib import Path
+from typing import Any
 
 import lsstypes as types
 import numpy as np
-import numpy.typing as npt
 import xarray as xr
 
+from .backends.jaxpower import JaxpowerBackend
```
epaillas, line 10: I would like to avoid the users having to install all three (and other possible future) backends unless necessary. How is this going to work with the optional dependencies?

epaillas:
I don't see any major blocks, so feel free to merge if you're happy with the changes. I ran it locally and though I'm getting some complaints, it's not strictly related to the PR, so let's just go ahead.

acm/estimators/galaxy_clustering/base.py
SBouchard01: This is necessary with the current implementation (lazy loading is implemented only in Python 3.14 IIRC and PEP8 imposes imports at the beginning of the file.

The correct approach would be to instead make a backend factory method (see #195, I've implemented something similar) to prevent this. I'd suggest to make a more general backend loading utils file later on which could allow us to load backend for any module - I'll open a new issue :)

SBouchard01:
> I don't see any major blocks, so feel free to merge if you're happy with the changes. I ran it locally and though I'm getting some complaints, it's not strictly related to the PR, so let's just go ahead.

What complaints do you get ? Are those related to some packages used ? Warnings for invalid-override-method are expected as most of our class implementations violate LSP - I'm solving this in #195 and will eventually do the same in other classes at some point)

SBouchard01:
Merging to update #195, let me know if the errors persist and if/how they can be solved ! :)

SBouchard01:
@epaillas I've added a backend registry utils class in #195 - once merged with `dev`, we can use that easily to register backends in `acm.estimators` without having to load all backends ai initialization:

```python
from acm.backend import BackendRegistry

_registry = BackendRegistry(BaseEstimatorBackend)
register_backend = _registry.register
load_backend = _registry.load

...

@register_backend("jaxpower")
class JaxPowerBackend(BaseEstimatorBackend)
    ...

class SomeEstimator:
    def __init__(backend: str):
        backend = load_backend(backend)
```

PR #244, opened 2026-04-22
Enhance EMC model training/optimization scripts and add support for transformers

epaillas:
This pull request introduces several improvements and bug fixes across the ACM observables and EMC pipeline, with a focus on enhanced model loading flexibility, improved plotting diagnostics, and more robust data handling. The most significant changes include support for multiple emulator model types, improved handling of unstable WST coefficients, and expanded plotting diagnostics for emulator residuals.

This goes in pair with [sunbird's PR68.](https://github.com/florpi/sunbird/pull/68)

**Model Loading and Flexibility:**

* Added support for loading different emulator model types (`FCN`, `Transformer`, `Zhong24Transformer`) by inspecting the checkpoint's metadata, allowing more flexible and future-proof model loading in `acm/observables/base.py`. The `load_model` method and related logic now dynamically select the appropriate class based on checkpoint contents. [[1]](diffhunk://#diff-4e8c04faed0419babbe4f3711d23c5600bd5a7481363d1ec29fd2386097dd972L14-R14) [[2]](diffhunk://#diff-4e8c04faed0419babbe4f3711d23c5600bd5a7481363d1ec29fd2386097dd972R29-R61) [[3]](diffhunk://#diff-4e8c04faed0419babbe4f3711d23c5600bd5a7481363d1ec29fd2386097dd972L213-R263) [[4]](diffhunk://#diff-4e8c04faed0419babbe4f3711d23c5600bd5a7481363d1ec29fd2386097dd972R277-R287)
* Updated type hints and method signatures throughout `acm/observables/base.py` to use `torch.nn.Module` instead of only `FCN`, reflecting support for multiple model architectures. [[1]](diffhunk://#diff-4e8c04faed0419babbe4f3711d23c5600bd5a7481363d1ec29fd2386097dd972L38-R71) [[2]](diffhunk://#diff-4e8c04faed0419babbe4f3711d23c5600bd5a7481363d1ec29fd2386097dd972L630-R669)

**WST Coefficient Masking:**

* Expanded the list of masked (unstable) WST coefficients in both `compress_covariance` and `compress_data` functions based on the latest tests by @gvalogia. [[1]](diffhunk://#diff-c9c887610359345bbad71c1689ab5ddb8d5b9b326121df66055783d5b58f0bfcR93-R99) [[2]](diffhunk://#diff-c9c887610359345bbad71c1689ab5ddb8d5b9b326121df66055783d5b58f0bfcR121-R139) [[3]](diffhunk://#diff-c9c887610359345bbad71c1689ab5ddb8d5b9b326121df66055783d5b58f0bfcL132-R160) [[4]](diffhunk://#diff-c9c887610359345bbad71c1689ab5ddb8d5b9b326121df66055783d5b58f0bfcR249-R255) [[5]](diffhunk://#diff-c9c887610359345bbad71c1689ab5ddb8d5b9b326121df66055783d5b58f0bfcR277-R295) [[6]](diffhunk://#diff-c9c887610359345bbad71c1689ab5ddb8d5b9b326121df66055783d5b58f0bfcL278-R332)

**Plotting and Diagnostics:**

* Enhanced the `plot_emulator_residuals` method to include a histogram of $\chi^2$ values, configurable binning/clipping, and improved layout. This provides better diagnostic insight into emulator performance and potential outliers. [[1]](diffhunk://#diff-4e8c04faed0419babbe4f3711d23c5600bd5a7481363d1ec29fd2386097dd972L989-R1030) [[2]](diffhunk://#diff-4e8c04faed0419babbe4f3711d23c5600bd5a7481363d1ec29fd2386097dd972L1003-R1065) [[3]](diffhunk://#diff-4e8c04faed0419babbe4f3711d23c5600bd5a7481363d1ec29fd2386097dd972R1084-R1120)
* Improved LaTeX formatting for plot labels in several EMC observable modules for consistency and clarity. [[1]](diffhunk://#diff-577e03e3c1efa3a6548180a5607f78113658c4d77e8b91670b29f39710239fe0L311-R311) [[2]](diffhunk://#diff-9c3e9d9d9ddba1be0ebb8f7e28cea21c33b65586a20fe8e6f03c25e84b61bb9dL309-R309) [[3]](diffhunk://#diff-4e03b4415b3943b8d6a311bb5485adb0b32a1e0fcfdd82b5ea0a87f8d4a545c6L350-R350) [[4]](diffhunk://#diff-b089f0e6910a9dbbee7bd4eb19f7bc9143b76767bdc1a3e12820786da0d8417aL354-R354)

**API Consistency and Bug Fixes:**

* Standardized the use of `ells` instead of `multipoles` in filter dictionaries throughout the codebase, fixing a bug and ensuring consistency. [[1]](diffhunk://#diff-577e03e3c1efa3a6548180a5607f78113658c4d77e8b91670b29f39710239fe0L292-R292) [[2]](diffhunk://#diff-bd59a0c8136ce61a0539d048fb0b4e72244173e2b17bb12c39ead1432eff9083L113-R120)
* Updated the inference script to support passing a model directory via the `--model_dir` argument, improving reproducibility and configurability for checkpoint selection. [[1]](diffhunk://#diff-bd59a0c8136ce61a0539d048fb0b4e72244173e2b17bb12c39ead1432eff9083R139) [[2]](diffhunk://#diff-bd59a0c8136ce61a0539d048fb0b4e72244173e2b17bb12c39ead1432eff9083R269)

**Other Improvements:**

* Minor import and type hinting adjustments for better code clarity and static analysis support.


epaillas:
I'll address the failed checks in a minute...

SBouchard01:
@epaillas would it be Ok to hold this until #241 is merged to avoid fixing twice the code typing/formatting issues ? (also that will leave me the time to review the PR tomorrow morning ^^)

acm/observables/emc/wst_module.py
```diff
@@ -111,6 +118,25 @@ def compress_covariance(
...
+            182,
+            184,
+            185,
+            186,
+            190,
+            195,
+            196,
+            197,
+            200,
+            201,
```
SBouchard01: At this point, I'd make those indexes a csv file that I would import as a class setting parameter instead of hardcoded magic values

acm/observables/emc/wst_module.py
```diff
@@ -241,6 +272,25 @@ def compress_data(
...
         ]
```
SBouchard01: Especially since those are the same indexes

acm/observables/base.py
```diff
@@ -24,6 +24,43 @@
 # Register safe globals for transform classes to allow loading checkpoints
 # with PyTorch 2.6+ (which changed weights_only default to True)
 SAFE_CLASSES = [LogTransform, ArcsinhTransform]
+EmulatorModel: TypeAlias = FCN | Transformer | Zhong24Transformer
+EmulatorModelClass: TypeAlias = type[FCN] | type[Transformer] | type[Zhong24Transformer]
```
SBouchard01: Not needed, as the line before is semantically the same in Python>3.10
see: https://typing.python.org/en/latest/spec/aliases.html

acm/observables/base.py
```diff
@@ -24,6 +24,43 @@
...
+    model_type = str(hparams.get("model_type", "fcn")).lower()
+    if model_type == "fcn":
+        return FCN
+    if model_type == "transformer":
+        return Transformer
+    if model_type == "zhong24_transformer":
+        return Zhong24Transformer
+    raise ValueError(
+        f"Unknown emulator model_type '{model_type}' in checkpoint {checkpoint_fn}."
+    )
```
SBouchard01: Outsourcing all of that seems a good idea, but IMO it would be better to have that in Sunbird and load the model trough a Sunbird method.

Also, this might move the issue, but any future implementations would require a code change - we might want to solve this by passing the model type as a class parameter somehow... Maybe a new method that would set the model class ?

Ps: is the first method used ??

acm/observables/base.py
```diff
@@ -35,7 +72,7 @@ def __init__(
         self,
         stat_name: str,
         dataset: xarray.Dataset | None = None,
-        model: FCN = None,
+        model: Optional[EmulatorModel] = None,
```
SBouchard01: Use `EmulatorModel | None` instead

acm/observables/base.py
```diff
@@ -126,16 +163,28 @@ def __init__(
...
+                    checkpoint_fn = Path(checkpoint_fn)
+                    self.model = self.load_model(checkpoint_fn)
+                except (
+                    FileNotFoundError,
+                    OSError,
+                    RuntimeError,
+                    KeyError,
+                    ValueError,
+                ) as e:
+                    logger.warning(f"Could not load model from checkpoint: {e}")
```
SBouchard01: Moving the try/except outside the "if paths is not None" loop prevents catching the model loading error with the newt model loading + duplicates code !

epaillas:
> @epaillas would it be Ok to hold this until #241 is merged to avoid fixing twice the code typing/formatting issues ? (also that will leave me the time to review the PR tomorrow morning ^^)

Sure!

acm/observables/base.py
```diff
@@ -659,11 +709,17 @@ def get_model_prediction(
...
-        if model is None:
-            model = self.model
+        if model is not None:
+            resolved_model = model
+        elif hasattr(self, "model"):
+            resolved_model = self.model
+        else:
+            raise AttributeError(
+                "No model loaded. Please provide a model or initialize the observable with model paths."
+            )
```
SBouchard01, line 650: This is a good idea 👍🏼

SBouchard01 (changes requested):
A few things to change in my opinion on the base Observable class

The rest looks good to me.

I didn't check the emc scripts I'll trust you to test if they are stillworking

acm/observables/base.py
epaillas: Which part specifically is not needed? Are you suggesting to only use EmulatorModel and drop line 28?

acm/observables/base.py
epaillas: The try/except is still there for the "paths is not None" branch in the latest commit. About the duplication, do you have any suggestion?

acm/observables/emc/wst_module.py
epaillas: Good call. I moved the WST mask indices into a packaged CSV and exposed them as a class setting in [42d99db](https://github.com/epaillas/acm/pull/244/commits/42d99db6a21f3b11e598fec4d14d84ca6bb2e27e)

acm/observables/emc/wst_module.py
epaillas: Good call. I moved the WST mask indices into a packaged CSV and exposed them as a class setting in [42d99db](https://github.com/epaillas/acm/pull/244/commits/42d99db6a21f3b11e598fec4d14d84ca6bb2e27e)

acm/observables/base.py
epaillas: Added in [d0cf3a7](https://github.com/epaillas/acm/pull/244/commits/d0cf3a77f4a3ccc7a56f164a46afd05496499355)

acm/observables/base.py
epaillas: Good point. I tried this in [99b31f1](https://github.com/epaillas/acm/pull/244/commits/99b31f1d15b7ed2c22613010d45868f4603ddf95), which goes together with sunbird's [PR 69](https://github.com/florpi/sunbird/pull/69).

acm/observables/emc/wst_module.py
SBouchard01: I think it would be best to have the CSV in the EMC scripts, as this is not dependent on the measurement method but on the mocks, and then pass this file as an optional mask argument to the class (e.g. move `acm/acm/observables/emc/data` to some folder in `acm/scripts/emc`). Ideally, I'd even recommend not having this data file at all in the repo and move it with the compressed files.

acm/observables/base.py
SBouchard01: Yes, in Python > 3.10 `type(SomeClass)` is semantically the same as the class itself for type hints, so line 28 is redundant with line 27

SBouchard01: Also, if only using the `BaseModel` Sunbird class, there is no need for a TypeAlias anymore

acm/observables/base.py
SBouchard01: New implementation is more consistent, loading the checkpoint_fn and then loading the model in a try/except block is OK (and less complex in terms of indentations)
I'd just change `checkpoint_to_load` by `checkpoint_fn` as it is basically the same object (but I'm nitpicking)

SBouchard01:
@epaillas I answered some of your comments :) 

PS: tests are failing because of the following error
```
ImportError: cannot import name 'BaseModel' from 'sunbird.emulators'
```

> Note to self: Are we sure this isn't breaking previous existing sunbird model loads ? Just to be sure...

acm/observables/base.py
SBouchard01: Erratum: this is actually useful for type hints, (the fisrt one is a class _instance_, the second one is a class _type_ see [this example](https://adamj.eu/tech/2021/05/16/python-type-hints-return-class-not-instance/)) so we can keep it. However, I'm not sure there is a big benefit in redefining a type alias for one specific type, we could just move that wherever it's used...

epaillas:
@SBouchard01 checks are passing now. Do you think this is in good shape to merge?

acm/observables/base.py
```diff
@@ -21,9 +19,8 @@
 
 logger = logging.getLogger(__name__)
 
-# Register safe globals for transform classes to allow loading checkpoints
-# with PyTorch 2.6+ (which changed weights_only default to True)
-SAFE_CLASSES = [LogTransform, ArcsinhTransform]
+type EmulatorModel = BaseModel
+type EmulatorModelClass = type[BaseModel]
```
SBouchard01: Type aliases are not that complex, and used at 3 places. We can directly use those in the type hints instead (it makes the code clearer IMO)

acm/observables/base.py
```diff
@@ -205,46 +214,27 @@ def load_dataset_from_files(cls, stat_name: str, paths: dict) -> xarray.Dataset:
...
-            logger.warning(
-                "Retrying checkpoint load with weights_only=False for %s due to PyTorch weights-only unpickling restrictions.",
-                checkpoint_fn,
-            )
-            model = FCN.load_from_checkpoint(
-                checkpoint_fn, strict=True, weights_only=False
-            )
-        model.eval().to("cpu")
-        return model
+        return load_model_from_checkpoint(checkpoint_fn, model_cls=model_cls)
```
SBouchard01: Since we now just return that method, do we want to keep the `load_model` classmethod ? (now we can just pass `class.model = load_model_from_checkpoint(checkpoint_fn, model_cls=model_cls)` 🤔 )

acm/observables/emc/wst_module.py
```diff
@@ -14,37 +14,17 @@
...
 
+    mask_indices = np.loadtxt(
+        Path(__file__).with_name("data") / "wst_mask_indices.csv",
+        delimiter=",",
+        skiprows=1,
+        dtype=int,
+    )
+
     def __init__(self, stat_name: str = "wst", **kwargs) -> None:
         super().__init__(stat_name=stat_name, **kwargs)
```
SBouchard01: I'd move that file to a local folder, and pass this as an initialization argument instead (it's best to avoid data files & magic values in the repo)

```suggestion
    def __init__(self, stat_name: str = "wst", mask_csv: str | Path | None = None, **kwargs) -> None:
        super().__init__(stat_name=stat_name, **kwargs)
        self.mask_indices = np.loadtxt(mask_csv, delimiter=",", skiprows=1, dtype=int) if mask_csv else []
```

tests/acm/observables/test_base_observable.py
SBouchard01, line 1: Could you add a small 1-line docstring to each new test to clarify what each test is doing ? :)

acm/observables/emc/wst_module.py
SBouchard01: + remove `acm/observables/data/wst_mask_indices.csv`

tests/acm/observables/test_base_observable.py
```diff
@@ -70,6 +69,138 @@
...
+
+    model = Observable.load_model(checkpoint_fn, model_cls=FakeModelClass)
+
+    assert isinstance(model, FakeLoadedModel)
+    assert calls == [
+        {
+            "checkpoint_fn": checkpoint_fn,
+            "model_cls": FakeModelClass,
+        }
+    ]
```
SBouchard01: > [!NOTE]
> If removing `observable.load_model` (as sunbird handles all the calls anyways), this test is no longer needed.
> Though, we could add this kind of test in sunbird instead...

SBouchard01 (changes requested):
Still a few comments (updated from last time. I really like the simplification of the model loading ! 

Tests are Ok too :) 

As a last suggestion, maybe you coud try to rebase the branch on dev before merging to cleanup the tree ? You might have to solve conflicts for the 3 observables files though, but I find the VSCode merge editor quite useful in solving those :)

epaillas:
@SBouchard01 see if you're happy with the latest changes.

SBouchard01 (approved):
Looks good to me !
Since `acm.observables.base` has been modified, I'd recommend rebasing on dev after merging #245 (because this branches from 0.3.0-dev10, #245 brings the version to 0.3.0-dev15) and ensuring tests pass but overall this PR is OK to merge 👍🏼 

> Still not checking the scripts - I trust you to have tested those ^^

PR #245, opened 2026-04-29
[Feature] Rework compression methods

### LINKED ISSUE #184: [Feature] Remove code duplication in the compress_data() method in the derived BaseObservable classes

### LINKED ISSUE #246: [Feature] Streamline Observable code
Some methods in Observable classes could gain in clarity by changing a few things:
- [ ] Move plot methods outside of the classes. I think most of those methods don't have a real gain at being inside the class, and could be moved in a dedicated `acm.observables.plot` module.
- [x] Try to refactor [methods with too many parameters](https://softwareengineering.stackexchange.com/questions/145055/are-there-guidelines-on-how-many-parameters-a-function-should-accept) in smaller chunks
- [x] Unmix properties, methods and dataset elements sharing the same names - this can cause typing issues and makes bug hunting more complicated (namely for `emulator_covariance_y` and `emulator_error`)
- [ ] Move projects control plots to dedicated methods in scripts instead of being in Observable classes. Those are often buggy due to the specific nature of such plots and the calls of the Observable classes. They could gain at being properly defined with the correct observable calls outside of the methods, and would be more relevant in scripts than in the package itself.
- [x] When relevant, move staticmethods outside of the classes (see [this comment](https://stackoverflow.com/a/11788267) explaining the reasoning with staticmethods vs functions). This includes loading the model (that could be done directly from `sunbird`?), `stack_on_attribute`, `flatten_output`
- [x] Deprecate compression methods in favor of compression scripts (see #245)

> [!NOTE]
> By deprecating compression methods and plot methods in Observable child classes, we reach a point where the only element remaining in most of those classes are the default init values and the emulator covariance error computation, both done in the base project classes. We could then deprecate all child observables classes in favor of calling only the main child class as e.g. `tpcf = EMCObservable(stat_name='tpcf')`)

_Side notes:_
- [ ] Replace elements like `paths if paths else {}` by `paths or {}` for clarity

SBouchard01:
Following #184, I introduce a new utils file that should lead to the deprecation of the Observables compression methods.

By using regex on glob search, this PR allows us to move the compression functions outside of the scope of each filetype and project

This files introduces 2 main methods: 
- `collect_mocks` to extract filenames from a glob pattern and a root directory, returns indices for each group of files found from the pattern. e.g. for `c{cosmo_idx}_ph{phase_idx}/hod{hod_idx}/tpcf_los_{los}.npy`, will return all filenames grouped by cosmo + phase + hod + los indices.
- `compress_mocks` to compress all mocks found by the previous method in one `DataArray`. It has a `reader` method and a `filter` method in input to not depend on the file type of the manipulation on those. The reader reads and merges each group of files in one object, the filter handles these objects and returns an array and a dictionary containing the _dimensions_ of this array (passed to `xarray`)

> [!NOTE]
> I'm not sure of the format of the groups, at the moment it's a dict with a tuple of indices as the key, and a list of filenames as the values, but it can be a bit complicated to handle. We could find an easier solution ?

I also added 2 temporary methods (`compress_x` and `compress_data`), adapted from the BGS file compression - that could eventually be moved somewhere else.
**TBD: Do we want those methods in acm or can we move those in the scripts ?**

### DIFF
--- a/acm/utils/compression.py
+++ b/acm/utils/compression.py
@@ -595,6 +595,14 @@ def compress_mocks(
         )
 
     sample_dims = {idx: np.unique(values) for idx, values in index_arrays.items()}
+    n_groups = len(groups)
+    n_expected = np.prod([len(v) for v in sample_dims.values()])
+    if n_groups != n_expected:
+        raise ValueError(
+            f"Index grid is sparse: found {n_groups} groups but expected {n_expected} "
+            f"from unique index combinations {({k: len(v) for k, v in sample_dims.items()})}. "
+            "Ensure all combinations of index values are present in the data."
+        )
 
     coords = cast_coords({**sample_dims, **features_coords})
     data = reshape_to_coords(selected_results, coords)
@@ -618,7 +626,7 @@ def compress_mocks(
 def compress_x(
     root_dir: str | Path,
     index_arrays: dict[str, list],
-) -> xarray.DataArray:
+) -> xarray.DataArray: # pragma: no cover
     """
     Compress "x" data from csv files based on collected index arrays.
 
@@ -687,7 +695,7 @@ def compress_data(
     override_last_dim: bool = True,
     save_fn: str | Path | None = None,
     **kwargs,
-) -> xarray.Dataset:
+) -> xarray.Dataset: # pragma: no cover
     """
     Compress mock measurement data into an xarray Dataset.
 
--- a/acm/utils/compression.py
+++ b/acm/utils/compression.py
@@ -613,6 +613,8 @@ def compress_mocks(
     )
 
     if drop_singleton_dims:
+        singleton_dims = [dim for dim in cout.dims if cout.sizes[dim] == 1]
+        logger.debug(f"Dropping singleton dimensions: {singleton_dims}")
         cout = cout.squeeze(drop=True)
 
     cout.attrs = {
--- a/acm/utils/compression.py
+++ b/acm/utils/compression.py
@@ -1,0 +1,738 @@
+import re
+import time
+import logging
+from typing import Any
+from pathlib import Path
+from collections.abc import Callable
+
+import xarray
+import lsstypes
+import numpy as np
+import pandas as pd
+from pycorr import TwoPointEstimator
+
+from acm.utils.xarray import split_vars, dataset_to_dict
+
+logger = logging.getLogger(__name__)
+
+#%% Readers and filters for different file formats
+def lsstypes_reader(files: list[Path]) -> Any:
+    """
+    Read and average a list of lsstypes files.
+
+    Parameters
+    ----------
+    files : list[Path]
+        List of file paths to read using lsstypes.
+
+    Returns
+    -------
+    Any
+        Averaged lsstypes object across all input files.
+    """
+    loaded = [lsstypes.read(f) for f in files]
+    data = lsstypes.mean(loaded)
+    return data
+
+def lsstypes_filter(
+    data: list[Any], 
+    last_dim: str, 
+    select: dict, 
+    get: dict, 
+    rebin: dict | None = None,
+) -> tuple[np.ndarray, dict]:
+    """
+    Apply selection, optional rebinning, and coordinate extraction to a list of lsstypes objects.
+
+    Filters all elements in ``data`` to match the shape and coordinates of the
+    first element after applying the selection and get operations, then stacks
+    the results into a numpy array.
+
+    Parameters
+    ----------
+    data : list[Any]
+        List of lsstypes objects to filter.
+    last_dim : str
+        Name of the last (feature) dimension, used to extract coordinates
+        from the filtered data and append to the returned dimension dict.
+    select : dict
+        Keyword arguments passed to ``.select()`` to slice the data along
+        one or more dimensions (e.g. ``{'k': (0.01, 0.7)}``).
+    get : dict
+        Keyword arguments passed to ``.get()`` to extract specific entries
+        along one or more dimensions (e.g. ``{'ells': [0, 2]}``).
+    rebin : dict, optional
+        If provided, keyword arguments passed to a first ``.select()`` call
+        before the main selection, used to downsample the data
+        (e.g. ``{'k': slice(0, None, 13)}``).
+
+    Returns
+    -------
+    data_out : np.ndarray
+        Array of shape ``(len(data), ...)`` containing the filtered and
+        matched data for each input object.
+    dims : dict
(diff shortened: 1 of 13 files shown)


SBouchard01:
This should also be easier to test the compression methods

acm/utils/compression.py
```diff
@@ -0,0 +1,789 @@
...
+            poles = d[q][::rebin](ells=ells, return_sep=False)  # ty:ignore[not-subscriptable]
+            data_out.append(poles)
+
+    # Stack the results into a numpy array
+    data_out = np.stack(data_out)
+
+    # Create the dimension dictionary
+    dims = {"quantiles": quantiles, "ells": ells, "s": s}
+
+    return data_out, dims
```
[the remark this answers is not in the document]
SBouchard01: TODO: ensure processors output match `(len(data) *dims)` shape

acm/utils/compression.py
```diff
@@ -0,0 +1,789 @@
...
+    for key in groups:
+        for idx, value in key:
+            index_arrays[idx].append(value)
+
+    logger.debug(
+        f"Found {len(files)} files grouped into {len(groups)} unique index combinations."
+    )
+
+    return groups, index_arrays
+
```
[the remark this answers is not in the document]
SBouchard01: TODO: Unit tests

acm/utils/compression.py
```diff
@@ -0,0 +1,789 @@
...
+    sample_dims = {idx: np.unique(values) for idx, values in index_arrays.items()}
+
+    dims = cast_dims({**sample_dims, **features_dims})
+    shape = [len(v) for v in dims.values()]
+
+    # TODO: drop singleton dims here instead ?
+    # Otherwise they still exist in the DataArray, which can be confusing
+
+    cout = xarray.DataArray(
+        data=selected_results.reshape(shape),
```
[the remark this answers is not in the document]
SBouchard01: Both cases should be verified when passing groups and index_array

SBouchard01:
The compression functions can now be easily created at the project level, without depending on the file or estimator type; see the example `compress_data` function constructor:

```python
groups, index_arrays = collect_mocks(
    root_dir = Path(paths["measurements_dir"]) / "base",
    glob_pattern="c{cosmo_idx}_ph{phase_idx}/seed{seed}/hod{hod_idx}/" + glob_fn, # project dependent
)
x = compress_x(
    root_dir=paths["param_dir"],
    index_arrays=index_arrays,
)
y = compress_mocks(
    groups=groups,
    index_arrays=index_arrays,
    reindex=["hod_idx"], # project dependent
    reindex_group_by=["cosmo_idx", "phase_idx", "seed"], # project dependent
    **kwargs, # Passed to the processing function
)
data_vars = {"x": x, "y": y}
# Same goes for the covariance - see the detailed function
cout = xarray.Dataset(data_vars=data_vars)
# Then, add test/train split and save dataset to disk
```

Subsequently, it is then easy to compress files trough e.g.:
```python
compress_data(
    glob_fn='power_spectrum*.h5', #filetype glob pattern
    paths=paths, # containing parameters and measurements paths
    test_filters=test_filters,
    save_fn = paths['data_dir'] + 'spectrum.npy', # where to save the compressed file
    last_dim = 'k', # From here, all args are passed to the processing function
    select = {'k': (K_MIN, K_MAX)}, 
    get = {'ells': [0, 2]},
    rebin = {'k': slice(0, None, 13)},
)
```

SBouchard01:
> [!NOTE]
> Waiting #241 to be merged to rebase on dev and open this PR to review

SBouchard01:
Added unit tests (with the help of Claude) that cover most of the edge cases of the compression functions. The only functions I didn't test are the reader/processor functions that require files to read
**Total PR coverage (ignoring example functions): 75%**

acm/utils/compression.py
```diff
@@ -0,0 +1,815 @@
...
+    files : list[Path]
+        List of file paths to read using lsstypes.
+
+    Returns
+    -------
+    Any
+        Averaged lsstypes object across all input files.
+    """
+    loaded = [lsstypes.read(f) for f in files]
+    data = lsstypes.mean(loaded)
```
epaillas, line 36: The reader in `compress_mocks` defaults to this. In what context are you averaging over the measurements?

acm/utils/compression.py
```diff
@@ -0,0 +1,815 @@
...
+            index_arrays[idx].append(value)
+
+    logger.debug(
+        f"Found {len(files)} files grouped into {len(groups)} unique index combinations."
+    )
+
+    return groups, index_arrays
+
+
+def compress_mocks(
```
epaillas: IMHO the methods names should be more general. We might want to reuse these utilities to compress actual data measurements, not just mocks. Plus, compress_mocks sounds like we are processing mock catalogs. Why not use "measurements" or something else instead?

epaillas (changes requested):

acm/utils/compression.py
SBouchard01: For lines of sight, I'm using one file per los on BGS (but otherwise, any other similar implementation)
One file found will make a 1-element list so it won't change anything

acm/utils/compression.py
SBouchard01: Sure, that's a good idea ! 
I made these utils in the context of mocks generation (hence the names) but you are right it's more generalizable :)

acm/utils/compression.py
SBouchard01: But overall any free field we want to average on (hence the `ignore_index` parameter in `collect_mocks`)

epaillas:
Do you mind if I push a folder with compression scripts for EMC to this PR?

SBouchard01:
If it does not conflict with the 2 files go ahead (i'm editing those to fix the changes requested ^^) You'll add those in scripts/emc ?

epaillas:
Yes, just a few scripts under scripts/emc

epaillas (approved):

SBouchard01:
Whoops, forgot to rebase on dev

PR #250, opened 2026-05-26
[Feature] Coverage tracker

### LINKED ISSUE #206: [Feature] Add coverage tracker
We could track the coverage percentage of the branches, and add a CI that raises an error on PRs if the new covrage number is lower than the former one. This would incite the creation of tests to _at least_ keep the coverage number identical.

### Implementation
@luckyjim suggested storing the covering number somewhere during the tests CI. We could fetch that number at the start of the PR from the original branch, and compare it to the new one in the new branch. 

### Aesthetics
We could also add a coverage shield in the README.md file that is updated (see for example https://github.com/marketplace/actions/code-coverage-summary)

### LINKED ISSUE #211: [Feature] Serve code coverage report in GitHub Actions
It could be nice to serve the tests coverage output somewhere (see for example https://github.com/ctrf-io/github-test-reporter, we could use this to have the coverage summary in the Action output)

### LINKED ISSUE #251: [Feature] Fix pyproject dependencies
The dependencies in the `pyproject.toml` and the architecture of the package could be slightly reworked to ensure a proper install and avoid a long install time (mostly for testing purposes)

**Key points:**
- The cosmodesi NERSC install requires us to remove packages from the `cosmodesi` env from the default packages to avoid some dependencies issues
- `sunbird` should be added to the core ACM dependencies.
- Should we leave the install of backend packages to the user? Technically, we only provide some implementation to load any packages. We could also add some optional dependency group packaging all the acm-provided backends for each backend group we have: estimators, dark matter (see #195 ), etc.
    - e.g. for estimators - such as `pycorr`, `kymatio`, `mistreeplus`, or `numba` for the knn - could be bundled in one `estimator` dependency group
- Tests should only require the base install. If possible, we could minimize the packages needed for tests by mocking a maximum of packages in the unit tests (such as `AbacusHOD` being mocked in the new `catalogs` module - see #195 )
    - Other tests can also have untested dependencies - if those can be loaded with minimal dependencies and then mocked (such as `cosmoprimo` or `pycorr`), this can make a significant gain in install time.
    - #250 implements a `minimal-test` optional group to solve this - could be moved as the main `testing` group eventually (I need to check with @luckyjim the dependencies we can remove from this group)

> [!NOTE]
> End-to-end tests will still probably require a full install anyways :/

- PolyBin3D
- getdist
- mockfactory (should be mostly deprecated with #195 - unless we decide to keep the hod module?)

SBouchard01:
Add a coverage tracker in a coverage.txt file, that is updated on pushes to `dev` (and adds a badge in the README)

The same workflow also publishes a single message on the Pull Request to provide the coverage report.

Source: https://github.com/marketplace/actions/pytest-coverage-comment


epaillas:
Excellent! This is a very useful addition. Looks like it is ready to merge?

SBouchard01:
I'm adding a few extra tests in the next commit, if they pass and you are OK with this PR I will merge it ^^

SBouchard01:
On a side note: This new comment on the PR should give people the incentive to add tests as they add code to the package (it raises an error if the coverage value goes down), we should also progressively add _unit_ tests for the parts of the code that are not covered yet - When we go over the parts of the code that have not been reworked to the new standards (mostly estimators and Observables subclasses) this number should go up :) 
Some other files could be ignored as they represent specific things that can't be tested (like plotting utils functions)

I'm going to add a commit to not count some files in the coverage - @epaillas should we include `acm.hod` in those ignored files (if it is to be replaced by `acm.catalogs` from #195 eventually ?)

epaillas:
Yes, we can include abacus.hod in those files to ignore it, but let's not deprecate it until we've made sure all of our scripts have been ported to the new interface.

SBouchard01:
Sure! That's why I made it a new module instead of a rework in #195. We need at least to ensure our scripts are updated and maybe wait that the cutsky code is ported to the new module anyways ^^

SBouchard01:
If those tests pass, I will merge the PR.

Remaining tests on `acm.utils` will be done with the updated changes in #195

PR #252, opened 2026-06-04
Fix double application of box offset in voxel_position

acasella27:
Fixes a coordinate centering bug in voxel_position() from jaxel_voids.py where the offset (boxcenter - boxsize/2) was applied twice in workflows using random catalogs (has_randoms=True), causing void centers to be offset from true density minima and corrupting void–galaxy cross-correlations. The transformation from mesh to physical coordinates is now applied exactly once, independent of geometry.

Also added from __future__ import annotations (PEP 563) to defer evaluation of type hints, preventing TypeError at import time from union types like str | Path | None.

### DIFF
--- a/acm/estimators/galaxy_clustering/jaxel_voids.py
+++ b/acm/estimators/galaxy_clustering/jaxel_voids.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 import logging
 import time
 from collections.abc import Iterator
--- a/acm/estimators/galaxy_clustering/jaxel_voids.py
+++ b/acm/estimators/galaxy_clustering/jaxel_voids.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import logging
 import time
 from collections.abc import Iterator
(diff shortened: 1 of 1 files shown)


SBouchard01:
From what I know future imports for annotations will be deprecated in future Python versions (IIRC at 3.13 end-of-life). What Python version are you running on? Theoretically Union types hints are allowed in the 3.12 expected version acm is supposed to run on if those types are defined before being called 🤔

acasella27:
@SBouchard01 I’m on Python 3.12.9. The issue was with `TwoPointCorrelationFunction` being a function rather than a type in the `save` function. My understanding is that the annotation `TwoPointCorrelationFunction | None` was being evaluated at import/class definition time, which led to the TypeError (`TypeError: unsupported operand type(s) for |: 'function' and 'NoneType'`).

I added `from __future__ import annotations` to defer evaluation of annotations, which avoided the import-time failure in this case. That said, I’m not fully sure whether this is the intended or cleanest fix here.

SBouchard01:
Ah yes, that's a type error, functions are not valid types types anyways so the future import was likely moving that under the rug ^^
What is likely expected is a `BaseTwoPointEstimator` child class from `pycorr`.

I'll make a hotfix PR in a few hours for some typing issues of the new ty version, I'll also fix that so the tests pass. You'll have to rebase the branch on dev once it's merged, but it should solve that and allow the tests to pass on this PR 👍🏼

SBouchard01:
@acasella27 is it OK if I take over this branch to rebase on `dev` & remove the `future` import (this should have been fixed in #253 ) before merging ?

acasella27:
@SBouchard01 Sounds good!

SBouchard01:
@acasella27 PR has been merged ! I'll let you delete the branch if you are not working on it anymore (to avoid stale branches)
Thank you for the fix ! :)

Issue #243, opened 2026-04-21
[Fix] Make estimator backends consistent


SBouchard01:
The fixes in #241 raised one small issue: the galaxy clustering estimator backends have some inconsistencies in typing and/or function calls & signature that can raise issues.

I've fixed some of them, but it's mostly a band-aid on a deeper issue. I'd recommend to slightly rework the backends so that: 
- They share common method names and signatures (so that any estimator call is independent of the backend)
- If an estimator is dependent on a specific backend, it should be enforced at initialization (see: CIC)

> [!NOTE]
> We could use this opportunity to fix some small details in the estimators (e.g. docstrings in prep. for the autodoc later, or some unit tests implementation for each observable)

SBouchard01:
Noting that the Minkowski estimator is not using the backend and could be upgraded

PR #253, opened 2026-06-05
[Hotfix] ty update & pyproject rework (#251)

### LINKED ISSUE #251: [Feature] Fix pyproject dependencies

SBouchard01:
This PR is a hotfix for the new type errors raised by the new ty update to allow tests to pass again.

I also fixed some extra type issues (including one raised in #252, where some functions used the `pycorr.TwoPointCorrelationFunction` function as a type hint - which caused issues. I fixed that by replacing those hints by `BaseTwoPointEstimator` hints instead.

I noticed that on a full install ty still raises some errors for galaxy estimators - to be fixed with #243

### DIFF
--- a/.github/workflows/ci-format-lint-test.yml
+++ b/.github/workflows/ci-format-lint-test.yml
@@ -123,7 +123,7 @@ jobs:
       - name: Installation package with editable mode
         run: |
           python -m pip install --upgrade pip
-          pip install -e .[testing,sunbird,cosmodesi,jaxpower]
+          pip install -e .[testing]
 
       - name: Process tests
         run: |
--- a/acm/estimators/galaxy_clustering/jaxel_voids.py
+++ b/acm/estimators/galaxy_clustering/jaxel_voids.py
@@ -290,7 +290,7 @@ class JaxelVoids(BaseEstimator):
     def save(
         self,
         filename: str | Path,
-        data: type[BaseTwoPointEstimator] | None = None,
+        data: BaseTwoPointEstimator | None = None,
         data_type: str = "catalog",
         attrs: dict | None = None,
     ) -> None:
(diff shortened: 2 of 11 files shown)


epaillas (approved):

epaillas:
Thanks for the changes. I don't see any blocking issues with this, so you can go ahead and merge.

PR #261, opened 2026-06-11
fix: try to fix the merge conflict on workflows 

SBouchard01:
… for PR merged to dev by checking out the repo on the commit ref instead of SHA


SBouchard01:
Merging assuming tests pass (they should) - I need to see if it works

Issue #43, opened 2025-10-20
[Feature] Update documentation


SBouchard01:
The `cosmodesi` branch contains a `docs` folder with an outdated doc that needs to be updated and filled with the ACM purpose and API.

It also needs to be hosted somewhere (TBD: Readthedocs, or epaillas.github.io/acm ? If we move this repo to cosmodesi, probably cosmodesi.github.io/acm ?)

epaillas:
@SBouchard01 Here's an attempt at hosting the documentation on readthedocs.io based on the draft PR above:

https://acm.readthedocs.io/en/latest/

There are parts of the API reference that are not building property, presumably due to import problems with Cython and other packages. For the moment, I have set the PR branch as the default in ReadTheDocs to test everything. Once we are happy with the implementation, we can merge the PR and I'll switch the default branch on the remote server.

epaillas:
> Yes, I'll need to rework the API calls to map the `cosmodesi` branch organization. In theory, since I mock import the Cython dependencies, the doc should compile cleanly (at least it was the case on the harmonized branch), but I'll check with the cosmodesi branch

Should we merge the PR to have the docs oficially hosted and made public and fix the problems in a separate Issue?

PR #266, opened 2026-06-16
[Refactor] Semantic changes in acm.catalogs (#264)

### LINKED ISSUE #264: [Feature] Semantic changes in acm.catalogs
Noting a few semantic changes for clarity in `acm.catalogs`:
1. `get_dark_matter_catalog` should be renamed `load_dark_matter_catalog` (get -> load, gives a better understanding of what the method does - loading data from disk)
2. catalog.positions should default raw to False, as this is an option and ideally getting catalog info should happen with transforms applied

### LINKED ISSUE #270: [Feature] Add deprecation warning to acm.hod
Now that `acm.catalogs` is working for snapshots, we can add a deprecation warning when loading `acm.hod`

Since the port of cutsky/lightcone functionalities is planned for  1.1.0, adding the deprecation warning to `acm.hod` in 1.0.0 makes sense ?

SBouchard01:
Fix some small semantic issues in `acm.catalogs`:
- Changed `get_dark_matter_catalog` to `load_dark_matter_catalog` (as it's read from disk) (#264)
- Added loading cache to the DM backends to not reload the dark matter catalog when calling make_catalog (which calls `load_dark_matter_catalog` at each run to get the different catalogs redshifts)
- Fixed a tracer name resolution issue in AbacusHODBackend that caused a crash
- Added deprecation warning in `acm.hod` (#270)
- Replaced default value raw=False in `SnapshotCatalog.positions` (#264)
- Added default import of available DM backends to make registration effective

### DIFF
--- a/acm/catalogs/backends/abacus.py
+++ b/acm/catalogs/backends/abacus.py
@@ -92,7 +92,7 @@ class AbacusHODBackend(SnapshotBackend):
         self.sim_type = sim_type
         self.sim_params = sim_params
         self.hod_params = hod_params
-        
+
         self._cache: dict[float, AbacusHOD]
 
     @override
@@ -104,10 +104,10 @@ class AbacusHODBackend(SnapshotBackend):
     ) -> AbacusHOD:
         if redshift in self._cache and no_cache is False:
             logger.debug(
-                f'Loaded dark matter catalog for redshift z={redshift:.3f} from cache.'
+                f"Loaded dark matter catalog for redshift z={redshift:.3f} from cache."
             )
             return self._cache[redshift]
-        
+
         sim_params = self.sim_params.copy()
         sim_params["z_mock"] = redshift
 
--- a/acm/catalogs/backends/abacus.py
+++ b/acm/catalogs/backends/abacus.py
@@ -56,6 +56,7 @@ class AbacusHODBackend(SnapshotBackend):
         ValueError
             If the config file is not found.
         """
+        super().__init__()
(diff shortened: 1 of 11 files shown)


SBouchard01:
@hbrincon we'll probably need to rebase the cutsky factory branch (again) - sorry about that

acm/catalogs/backends/abacus.py
```diff
@@ -92,12 +93,21 @@ def __init__(
         self.sim_params = sim_params
         self.hod_params = hod_params
 
+        self._cache: dict[float, AbacusHOD]
+
     @override
-    def get_dark_matter_catalog(
+    def load_dark_matter_catalog(
```
hbrincon, line 99: Should there be a docstring here?

hbrincon (commented):
Looks good, only one small request for a change

acm/catalogs/backends/abacus.py
SBouchard01: I'm following [this principle](https://google.github.io/styleguide/pyguide.html#3831-overridden-methods), as it's an overridden method that matches the signature of the original, so the docstring is set in the base class (and it's also understood that way by type checkers: 
image
If (when fixing the docs in #43 ) we need to add the docstring here I can add it in a hotfix ? :)

hbrincon: Ok, sounds good as is then

hbrincon (commented):
Ready to merge

hbrincon (approved):

PR #267, opened 2026-06-17
[Feature] add script utilities

SBouchard01:
Here are some utilities I made while reworking the BGS scripts. I added them in the utils module of the acm package, let me know if you think it can be worth adding to the package :) 

You can see an example on how to use them on the new [BGS box measurement script](https://github.com/epaillas/acm/blob/scripts/bgs-v2/scripts/bgs/measurements/measure_box.py) (still WIP)


SBouchard01:
Note: I added a small typo fix to the coverage workflow to remove duplicate % sign if the coverage lowers

gvalogia (approved):

PR #274, opened 2026-06-23
[Feature] Estimator backends & common structure

### LINKED ISSUE #150: Update `BaseEstimator` child classes
Due to the changes made in #99, #144 and #149, some `BaseEstimator` child classes are now outdated and either need to be updated or removed (e.g. the `polybin3d` classes are now handled by the `jaxpower` backend of `BaseEstimator`), as they will now crash.

_The most obvious ones being_: `CountsInCells`, `MinkowskiFunctionals`, `MinimumSpanningTree`, and all the `polybin3d` classes

While doing this work, we could also consider improving some inheritance properties of the child classes (such as defining a common logger in `BaseEstimator` with `logging.getLogger(self.__class__.__name__)`) and avoiding writing by hand class properties that can be obtained trough calls to the class attributes

### LINKED ISSUE #179: [suggestion] use lsstypes as the default storage convention
For consistency between estimators, and loading within the Observable compression functions, we could use `lsstypes` objects as a storage convention.

see: https://github.com/adematti/lsstypes 

> [!NOTE]
> This could also allow us to eventually create some sort of factory/default method for the compression of observables ? This could help maintaining all the Observables classes

> [!TIP]
> `lsstypes` already has some conversion functions for some statistics (`pycorr`, `pypower`, ...), so we won't need to recompute all statistics !

### LINKED ISSUE #243: [Fix] Make estimator backends consistent

### LINKED ISSUE #268: [Feature] Flag outdated estimator classes
Once the base Estimator classes have been updated, we should flag the remaining estimator classes with a warning on call until they have been updated to the standard (to flag potential bugs sources)

### LINKED ISSUE #273: [Feature] Streamline compression methods
The current implementation (`acm.utils.compression`) is a bit convoluted and I think it could be solved more elegantly.

Also, the lsstype reader can't read some `lsstype` outputs (like outputs of `from_pycorr`) as they don't have a `mean()` property.

**Here are some ideas, to think about:**
1. use a dataclass instead of some dictionary with tuples to store the files indexes and paths
2. create a Compressor class with some inputs common to all measurements, and that has a compress method to return the compressed xarray
    - How to separate those informations properly and ensure indexation is consistent ?
    - We should ensure we don't encode project-dependent stuff there
    - Instead of having reader/processor methods, we make those internal class helpers: the user can create its own compresser by creating a subclass with custom helpers

SBouchard01:
> [!CAUTION]
> Delegated to #277 and  #278 - Keeping this PR open to track changes to this branch until all is ported to #278

This PR updates the galaxy estimator and their backends to a common standard:

- [x] Standardized backends
- [x] Added tests for backend functionalities 
- [x] Updated tests syntax to match some ruff & ty rules (because the warnings were annoying)
- [x] Reworked the base Estimator class with backend support & standard methods
    - Note that the backend attributes are now explicit to ensure consistency between backends !
- [x] Reworked some estimators as examples
- [x] Added unit tests for reworked estimators
- [x] Updated example notebooks for updated estimators
- [x] Tested consistency of reworked backend with previous results
- [x] Added warnings to non-updated estimators
    - To allow the tests to pass I disabled type checking _on purpose_ in the non-updated estimators. We expect the updates to remove the logging message and the `# type: ignore` on top of the file when ready.

> [!NOTE]
> The backend unit tests are done with mocked imports, which means they don't check the output and the inputs are dummy values.
> We will need to confirm the output are actually consistent and don't crash before merging !!


acm/estimators/galaxy_clustering/backends/jaxpower.py
```diff
@@ -285,40 +288,14 @@ def get_query_positions(
...
-        Returns
-        -------
-        array_like
-            Gaussian kernel in Fourier space.
-        """
-        return jnp.exp(
-            -0.5
-            * sum((kk * smoothing_radius) ** 2 for kk in mattrs.kcoords(sparse=True))
-        )
+        return np.asarray(coords, dtype=np.float32)  # NOTE: float32 mandatory here ?
```
SBouchard01: @epaillas I've kept the `np.float32` dtype, but is it necessary to hardcode it here ? 🤔

acm/estimators/galaxy_clustering/backends/pypower.py
```diff
@@ -17,157 +22,207 @@ class PypowerBackend:
...
-            # shift = self.mesh.boxsize / 2 - self.mesh.boxcenter
-            self.randoms_mesh = randoms_mesh
         else:
-            self.mean = np.mean(data_mesh)
-            delta_mesh = data_mesh / self.mean - 1
-        self.data_mesh = data_mesh
-        self.delta_mesh = delta_mesh
+            delta_mesh = data_mesh / np.mean(data_mesh) - 1
+
+        self._density_contrast = delta_mesh
```
SBouchard01, line 167: @epaillas do you know the type of that object ? The old `set_density_contrast` method was returning it as `array_like` but it has a `readout` method, so I'm not exactly sure what `delta_mesh` is

acm/estimators/galaxy_clustering/backends/pyrecon.py
```diff
@@ -1,279 +1,211 @@
...
+            randoms_weights,
+        )
+
+        # Set private attributes
+        self._boxsize = _make_array(boxsize, 3)
+        self._boxcenter = _make_array(boxcenter, 3)
+        self._meshsize = _make_array(meshsize, 3, dtype=int)
+
+        data_mesh = RealMesh(boxsize=boxsize, boxcenter=boxcenter, nmesh=meshsize)
+        data_mesh.assign_cic(data_positions, data_weights, wrap=True)
```
SBouchard01, line 53: > [!NOTE]
> Pyrecon needs positions to paint the particles (while other backends store them). I did not store the positions in the backends for memory purposes, as they exist in the code that calls the backend and are often also stored in the backend modules/
> This means I have to paint the particles in the initialization of the class here...

SBouchard01:
Also I've moved `_make_array` (that I have taken from pyrecon and updated) to `acm.utils.default`, I was not sure where to put that helper function :)

acm/estimators/galaxy_clustering/backends/pypower.py
```diff
@@ -17,157 +22,207 @@ class PypowerBackend:
...
+        np.ndarray
+            Density contrast at the input positions.
+        """
+        if self._density_contrast is None:
+            raise AttributeError(
+                "Density contrast has not been set, run set_density_contrast first."
+            )
+        t0 = time.time()
+        # offset = self.boxcenter - self.boxsize / 2.0
+        # positions = positions - offset
```
SBouchard01: @epaillas I have commented out this offset for now as it seems arbitrary compared to other backends.
What is the reason for its presence ? and do we need it here or is it an artefact from older versions ?

acm/estimators/galaxy_clustering/backends/jaxpower.py
epaillas: It's been a while since I implemented this... but sometimes these were needed to avoid issues with JAX across different machines...

I guess we could drop it in the future and see what happens, but I would leave it for now until we check that we reproduce the output of the previous code.

acm/estimators/galaxy_clustering/backends/pypower.py
epaillas: Probably a pmesh ParticleMesh object...

https://github.com/cosmodesi/pypower/blob/8606a05b357357c7cc71a5324cc3fdbdda522875/pypower/mesh.py#L10

acm/estimators/galaxy_clustering/backends/pyrecon.py
epaillas: Yes -- Usually with this backend, the scripts that call the estimator would run the assign_data and assign_random methods. The nice thing about that was that you could paint one random catalog at a time, without the need of storing all 18 DESI randoms to memory simultaneously.

The other backends don't allow that so I'm fine with deprecating that option, I guess...

acm/estimators/galaxy_clustering/backends/pypower.py
epaillas: You are not going to get the correct results if you comment it out. It's a pmesh convention that real-space positions are wrapped at nmesh / 2. I.e. [0, ... boxsize / 2, - boxsize / 2, ...] --- like Fourier modes. If you don't add this offset, the power will be completely off. We could add a small comment for future reference.

acm/estimators/galaxy_clustering/backends/pyrecon.py
SBouchard01: It's technically still possible with this new backend, by re-running assign_cic on self.randoms_mesh (and adding the length of the extra positions to self._size_randoms
```python
backend = PyreconBackend(some_data, boxsize, meshsize, some_randoms)
backend.randoms_mesh.assign_cic(some_other_randoms)
backend._randoms_size += len(some_other_randoms)
```

I did not think of that because the other backends were not allowing that and I wanted a common use for all 3 backends.

I can implement back a private painting method in that backend if you want - but we would still need to have the same behavior as the other backends and this method should only extend this behavior.

SBouchard01: --> Added back as a method extending the `PyreconBackend` behavior in a81f5cb2cbe33b6c15e7462efdd3607654bab037

acm/estimators/galaxy_clustering/backends/pypower.py
SBouchard01: OK, I see. Yes, documenting this is a good idea. But we also need a check before applying offsets, as we don't assume any coordinate convention except "cartesian coordinates in shape (N, 3)" in the other backends. This offset assumes we recieve the positions between 0 and boxsize

Do we need to make such assumption (and therefore check this at init) in other backends ?

acm/estimators/galaxy_clustering/backends/pypower.py
SBouchard01: OK thank you ! 
I was not sure because the operation of different types with matrices can have strange behaviors.
Since pmesh is not an ACM dependency, I'm going to leave the type as Any for now - hopefully this does not raises any future issues (it should not as it is a private attribute)

SBouchard01:
I spent a bit of time resolving dependencies issues in the tests. To summarize, mocking non-installed dependencies only works if those dependencies are not attempted to be imported before mocking them.

Adding the `check_installed` function to import/register backends when their dependencies are installed seemed a good idea, but when some of those are installed in the test env (e.g. `jaxpower` in my case), it was imported before I being mocked in the tests, so the actual package was used and tests crashed against the mocked data.

To ensure the tests run consistently in all environments, I removed all `__init__` imports that depend on optional dependencies. That means the backends or estimators will have to be imported from their file before we are able to use those - this sacrifices some user-side ease to have more control over how the code behaves.

SBouchard01:
Closing, delegated to https://github.com/epaillas/acm/pull/277 and https://github.com/epaillas/acm/pull/278

PR #279, opened 2026-07-15
Feature/cutsky factory

### LINKED ISSUE #265: [Feature] Add `acm.hod` functionalities in `acm.catalogs`
`acm.catalogs` is still missing the cutsky and lightcone generation to be fully able to replace `acm.hod`

hbrincon:
Draft PR for comments

### DIFF
--- a/acm/catalogs/factories/cutsky.py
+++ b/acm/catalogs/factories/cutsky.py
@@ -34,6 +34,11 @@ class BaseCutskyFactory(BaseCatalogFactory):
         self.backend: SnapshotBackend
         self.catalog_class: type[CutskyCatalog]
         self._catalogs: dict[tuple[float, float], CutskyCatalog]
+        # Type checks
+        if not isinstance(self.backend, SnapshotBackend):
+            backend_type = type(self.backend)
+            error_message = f'The provided backend must be an instance of SnapshotBackend. The received type was {backend_type}'
+            raise TypeError(error_message)
 
     @abstractmethod
     def make_catalogs(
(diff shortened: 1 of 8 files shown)


acm/catalogs/products/cutsky.py
SBouchard01, line 197: We could use [native cache properties](https://docs.python.org/3/faq/programming.html#faq-cache-method-calls) instead ?

acm/catalogs/factories/cutsky.py
```diff
@@ -24,16 +29,25 @@ class BaseCutskyFactory(BaseCatalogFactory):
...
         **kwargs,
     ) -> None:
         super().__init__(backend, catalog_class, cosmo, cosmo_fid, **kwargs)
         # Type hints
         self.backend: SnapshotBackend
-        self.catalog_class: type[CutskyCatalog]
-        self._catalogs: dict[tuple[float, float], CutskyCatalog]
+        self.catalog_class: type[SnapshotCatalog]
+        self._catalogs: dict[tuple[float, float], SnapshotCatalog]
+        self.cutsky_catalog_class : type[CutskyCatalog],
```
SBouchard01: `cutsky_catalog_class` is hinted here but not set by the user ?
I'd use `catalog_class` to define the class that needs to be set for the final output (so `CutskyCatalog` instead) - `SnapshotCatalog` can always be set internally ?

acm/catalogs/factories/cutsky.py
```diff
@@ -74,6 +88,46 @@ def get_catalog(self, redshift_range: tuple[float, float]) -> CutskyCatalog:
...
+        Parameters
+        ----------
+        path : str | Path
+            Output directory. Created if it does not exist.
+        """
+        path = Path(path)
+        path.mkdir(parents=True, exist_ok=True)
+        for zranges, catalog in self._catalogs.items():
+            zlow, zhigh = zranges
+            catalog.save(path / f"catalog_z{zlow:.3f}_z{zhigh:.3f}.h5")
```
SBouchard01: suggestion: maybe use `catalog_z{zlow:.3f}-{zhigh:.3f}.h5` ?

acm/catalogs/factories/cutsky.py
```diff
@@ -93,23 +147,73 @@ def make_catalogs(
...
+            )
+            shifts = self.get_box_shifts(pos_min, pos_max, boxsize)
+            # doesn't make sense to have snapshotcatalog as input (multitracer) while having dataframe as output (all tracers forced into single dataframe)
+            replications = self.get_box_replications(
+                snapshot_catalog,
+                pos_min,
+                pos_max,
+                shifts=shifts,
+                distance_limits = (distance_limits[0], distance_limits[1]),
+            )
```
SBouchard01: Question 1: could those be used as external helpers or staticmethods ? (rule of thumb is that if you don't need `self` in the method, it should be OK to move it as a static method or an external function)

Question 2: could we move the tiling functions to SnapshotCatalog (either as  staticmethods or helpers ?)

acm/catalogs/products/cutsky.py
hbrincon: I didn't write this code, so I'm less familiar with what it's doing. My understanding is that `ngal` and `_transform_state` both return mutable values, so would we need to define `__eq__` and `__hash__` for `BaseGalaxyCatalog` and use `@lru_cache` for the two methods?

acm/catalogs/factories/cutsky.py
hbrincon: `self` is used in both of these methods though, or maybe I'm misunderstanding the request. What did you mean?

I lean towards keeping the tiling in cutsky since it is really part of defining the cutsky geometry, but I could be convinced otherwise (since you're the one developing the snapshot classes, I'll let you implement that if you like)

acm/catalogs/factories/cutsky.py
hbrincon: I've switched the two

acm/catalogs/factories/cutsky.py
hbrincon: Done

PR #281, opened 2026-07-24
[Feature] Split caching of dark matter catalogs in backends 

SBouchard01:
This PR makes the caching of the dark matter catalog in the backends explicit by splitting the `load_dark_matter` method in `load_dark_matter` and `get_dark_matter`. It changes the call of the factories `make_catalog` to use the `get_dark_matter` method instead, avoiding silent loads of dark matter catalogs in the `make_catalog` method, and allowing to pre-load dark matter catalogs in the backend before the factory calls.

> [!NOTE]
> As usual when allowing more freedom in the calls, this means we add an extra step in the existing calls, where the user will need to either call `backend.load_dark_matter` or `factory.backend.load_dark_matter` before calling `factory.make_catalog`...

This PR also removes redundant "abacus" word in `acm.utils.abacus` functions and moved the `check_catalog` method to the snapshot product file, as `boundary_check` for a clearer method scope.

I am also adding a change to the save/load methods of the galaxy catalogs, to now save the Cosmology objects as datasets of the h5 files - using borrowed methods from `cosmoprimo`.


acm/catalogs/products/base.py
```diff
@@ -203,6 +204,12 @@ def save(self, path: str | Path, *columns: str) -> None:
             )
             self._save_attrs(f)  # subclass-specific attributes
 
+            # Save cosmology objects as groups to preserve their state
+            g = f.create_group("cosmo")
```
hbrincon: I'm used to `f` being used to mean `file`, but for `g` we might want to use a more informative name like `group` (I'm not familiar with what the standard is here, but it would be easier to read)

acm/catalogs/products/snapshot.py
```diff
@@ -370,3 +370,68 @@ def _random_positions(
...
+        If True, positions are required to be in the range [-L_i/2, L_i/2) for each axis.
+        If False, positions should be in [0, L_i). Default is False.
+
+    Raises
+    ------
+    ValueError
+        If any of the positions fall outside the specified box boundaries.
+    """
+    # Pick precision
+    _dtype = np.float32 if check_in_float32 else np.float64
```
hbrincon: Why do we set the precision manually rather than allowing the user input to determine it? (since the later case is our usual method, e.g. https://github.com/epaillas/acm/blob/20432a558aec852b0a8ebec02a13f7e4608739f9/acm/catalogs/factories/cutsky.py#L295 )

acm/catalogs/products/snapshot.py
```diff
@@ -370,3 +370,68 @@ def _random_positions(
...
+    if len(boxsize) == 1:
+        boxsize = np.repeat(boxsize, 3)
+    elif len(boxsize) != 3:
+        raise ValueError(
+            f"boxsize should be a float or an array of shape (3,), but got {boxsize.shape}"
+        )
+
+    # Pick right and left edges for each dimension
+    offset = boxsize / 2 if center_at_zero else 0.0
+    L = np.array([0.0, 0.0, 0.0], dtype=_dtype) - offset
```
hbrincon: The variable names here are clear enough, but something like `left_boundary` could still be more descriptive

acm/utils/abacus.py
```diff
@@ -27,7 +27,7 @@
 }
 
 
-def get_abacus_simname(sim_type: str, cosmo_idx: int, phase_idx: int) -> str:
+def get_simname(sim_type: str, cosmo_idx: int, phase_idx: int) -> str:
     """Build the Abacus simulation name based on the provided parameters."""
```
hbrincon, line 31: Remove Abacus from docstring

acm/catalogs/backends/abacus.py
```diff
@@ -120,7 +123,33 @@ def load_dark_matter_catalog(
             f"Loaded dark matter catalog for redshift z={redshift:.3f} in {time.time() - t0:.2f} seconds."
         )
         self._cache[redshift] = dark_matter_catalog
-        return dark_matter_catalog
+
+    @override
+    def get_dark_matter_catalog(self, redshift: float) -> AbacusHOD:
```
hbrincon, line 128: Should this return AbacusHOD or something more generic so that different sims can be used?

acm/utils/abacus.py
hbrincon, line 91: Remove Abacus from docstring

hbrincon, line 98: Remove Abacus from docstring

hbrincon, line 141: Is this Abacus format still expected now that we are making the code applicable for different sims?

hbrincon, line 157: Remove hardcoding?

acm/utils/h5.py
```diff
@@ -0,0 +1,143 @@
...
+    obj : object
+        The object to convert.
+
+    Returns
+    -------
+    object
+        The JSON-serializable representation of obj.
+    """
+    if isinstance(obj, np.ndarray):
+        return {"__ndarray__": obj.tolist(), "__dtype__": str(obj.dtype)}  # ty:ignore[no-matching-overload] ???
```
hbrincon, line 42: Is the comment here referring to a resolved issue, or is there a problem?

acm/utils/h5.py
```diff
@@ -0,0 +1,143 @@
...
+    Returns
+    -------
+    object
+        The restored object.
+    """
+    if isinstance(obj, dict):
+        if set(obj.keys()) == {"__ndarray__", "__dtype__"}:
+            _object = obj.get("__ndarray__")
+            dtype = obj.get("__dtype__")
+            return np.array(_object, dtype=dtype)  # ty:ignore[no-matching-overload]
```
hbrincon, line 86: Likewise here

hbrincon (changes requested):
This PR is near ready to be pushed with a few potential changes. I've not reviewed the test files in detail but will trust that they have been ran and found to work as intended.

acm/catalogs/products/snapshot.py
SBouchard01: It's a legacy code but you are correct, I will change that :)

acm/catalogs/backends/abacus.py
SBouchard01: The `AbacusHODBackend` only handles `AbacusHOD` objects, other sims will be handled trough their own backend (or `HODDIES` ^^)

SBouchard01: (Also, this typing is more restrictive for subclasses which should be OK I guess ? I'm always a bit confused with LSP violations lol)

acm/utils/abacus.py
SBouchard01: It's still Abacus-related, I just removed abacus from the name as it is redundant with the module name (abacus.py)

acm/utils/abacus.py
SBouchard01: Idem

SBouchard01:
**Re: Abacus related methods**
@hbrincon 

It's not that I wanted to make this adaptable for other sims, just that the "abacus" word in methods was redundant with the module name (`abacus.py`), but they are still Abacus-related (hence the hardcoded structure, simulation name and documentations)

e.g. `acm.utils.abacus.get_abacus_simname` --> `acm.utils.abacus.get_simname` is simpler :) 

For the type: ignore issues, those stem from typing issues between `object` types and `np.ndarray` stuff. Not sure how to solve that, I think it's a ty bug so I commented it out 😅

hbrincon:
@SBouchard01 Got it, I've resolved the relevant conversations

acm/catalogs/products/snapshot.py
SBouchard01: Fixed in [26ebde1](https://github.com/epaillas/acm/pull/281/commits/26ebde1b6d8135732956afb18c4fa900d16997d8)

acm/catalogs/products/snapshot.py
SBouchard01: Fixed in [94568e0](https://github.com/epaillas/acm/pull/281/commits/94568e0dd0fba58da28a1d0eb01cc7a2d5d68d65), we can now set the array dtype (and precision) manually, defaulting with float64

acm/catalogs/products/base.py
SBouchard01: Fixed in [26ebde1](https://github.com/epaillas/acm/pull/281/commits/26ebde1b6d8135732956afb18c4fa900d16997d8)
I used the same convention as below, calling those `grp` instead of `g` (those are internal and directly defined from an explicit method, so it should be clear enough in context)

hbrincon (approved):
The PR is ready to merge

PR #282, opened 2026-07-28
[Hotfix] Make deepcopy of default dicts when updating the tracers

SBouchard01:
This is a small fix that could cause eventual issues when setting the same tracer for several snapshots.
I also added extra tests and clarified some docstrings on the tracer flags behaviors


hbrincon (approved):
These changes are good to be merged.

PR #289, opened 2026-09-01
hotfix: solve some dependencies issues (#285)

### LINKED ISSUE #285: [Bug] Dependencies issues
Two small dependencies issues I encountered:

1. `h5py` should be included in the testing dependencies, as we now heavily depend on it for saving files. It is technically included in the `sunbird` dependencies, but when `sunbird` is not installed, it will be missing.
2. On Windows (conda?) with intel processors, `healpy` installs `llvm-openmp` which reverts `intel-openmp` to a very low version, causing several issues - including Jupyter kernel crashes when showing plots. A quick-fix is to override the `intel-openmp` install to the latest version by hand (though I recommend using WSL instead) - this should be documented somewhere. See https://github.com/ContinuumIO/anaconda-issues/issues/13146 for more information

SBouchard01:
Fix `jax` dependency requirement in `acm.utils.scripts` by making `jax.clear_caches` optional if `jax` is not avaiable.

While technically always available due to `sunbird` dependencies, it is now clear that it is not a required `acm` dependency.

Also updating the `README.md`:
- Add a note in the readme related to `healpy` install on windows
- Update dependency info
- Add the information & contributions instructions for the test suite

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,30 +30,25 @@ pip install -e .[sunbird]
 > The `-e` flag is used to install the package in editable mode, which allows you to make changes to the code and have them reflected without reinstalling the package.
 
 ### Requirements
-Strict requirements are:
-- `numpy`
-- `scipy`
-- `pandas` 
-- `pyyaml`
-- `matplotlib`
-- `getdist`
+Requirements are handled in the `pyproject.toml` file. Notable default requirements include:
+- [`sunbird`](https://github.com/florpi/sunbird) to load and run the emulators
+- [`lsstypes`](https://github.com/adematti/lsstypes) to read and write statistics measurements
 
 > [!WARNING]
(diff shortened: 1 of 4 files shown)


czymh (commented):
Nice cleanup! Only one question: `healpy` is moved out of the base requirements, but `acm/lensing/abacuslensing.py` and `acm/hod/cutsky.py` still import it at module level. Should we keep it in base?

SBouchard01:
The idea of the `cosmodesi` dependencies is to avoid reinstalling complex stuff that already exist/is maintained in our NERSC environment (as there is no flag to _not_ install dependencies in the pip install system :/ ).

As `healpy` already exists at NERSC I removed it from the default install (which conveniently solves my issue of Windows/Intel install as I only install the base dependencies when developing on this system). 

if this eventually becomes an issue, we can always add it back to the default deps later on ^^

PR #290, opened 2026-09-03
feat: move LHC sampler in utils (#272)

### LINKED ISSUE #272: [Feature] Move `acm.hod.parameters` to another module
When we deprecate `acm.hod` in v1.1.0, we still need the parameters.py functions to sample the HOD parameters in a LatinHypercube.

I propose we move that file in `acm.utils` - maybe as `acm.utils.hod` ?

> [!NOTE] 
> This can also be the occasion to fix semantic issues in the implementations; like the `self.params` value being rewritten at several places - which can lead to unexpected results. 
> **This should not change API or results.** _(And we could also add tests)_

SBouchard01:
From #272, moving `acm.hod.parameters` to `acm.utils` to plan for `acm.hod` future deprecation.

This PR also reworks the structure of the LHC sampler class to have a better tracking of the datasets, and streamlines some features.

Adding tests and an example notebook.


czymh (commented):
The updated `save` method now raises a KeyError for missing columns in order, whereas the original implementation silently ignored missing keys. I’m not sure if this is intentional. If this is the intended behavior, I think these changes are ready to be merged.

SBouchard01:
> The updated `save` method now raises a KeyError for missing columns in order, whereas the original implementation silently ignored missing keys. I’m not sure if this is intentional. If this is the intended behavior, I think these changes are ready to be merged.

Not exactly intentional but it makes a clearer use case IMO (e.g. only provide the order you want, not an arbitrary columns, which could return a different number of columns than expected)

PR #291, opened 2026-09-09
fix: pytest update

SBouchard01:
In this PR:
- Update some old tests format to the common standard for readability
- Removing old outdated files
- Using [append import mode for pytest](https://docs.pytest.org/en/stable/explanation/pythonpath.html) as the test suite starts to be longer, it allows identical file/test names across different folders.


SBouchard01:
Nevermind the copilot review, it was a misclick on my side lol

czymh (approved):
Deleting the three estimator tests is fine, since PR #278  adds proper replacements under `tests/acm/estimators/`.

PR #293, opened 2026-09-09
feat: Add benchmark timer

SBouchard01:
Adding a small class that allows me to benchmark different steps I loop over in the BGS scripts, I think this can be useful somewhere else

Also moving the memory cleanup to a dedicated function to avoid recreating it in my scripts, and add the option to not clear jax caches.


czymh (approved):
Clean, well-tested utility PR. No blockers.

